PR #2, opened 2019-10-04
fix slow consumer duplicates

gcolliso:


gcolliso:
n/a after moving from legacy to current gitbook

PR #7, opened 2019-11-13
add whats new doc

gcolliso:
update menu
add what's new doc
rename what's new in 2.0 -> NATS 2.0

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -1,7 +1,8 @@
 # Table of contents
 
 * [Introduction](README.md)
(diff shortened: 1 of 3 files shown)


gcolliso:
I just used the summary info from the tech preview blog post. Add/update as necessary

gcolliso:
I agree, I'm not sure where they should go tho

derekcollison:
Response permission should be straightforward. The latency tracking will need to have its own section somewhere at some point.

PR #18, opened 2019-12-18
structured data: Go: handle decode errors

philpennock:
The automatic handling of encoding/decoding of structured data can fail, if the data is not in the expected format.
For Go, show how this is handled with a connection error handler option.

### DIFF
--- a/developing-with-nats/receiving/structure.md
+++ b/developing-with-nats/receiving/structure.md
@@ -7,7 +7,10 @@ For example, to receive JSON you could do:
 {% tabs %}
 {% tab title="Go" %}
 ```go
-nc, err := nats.Connect("demo.nats.io")
+nc, err := nats.Connect("demo.nats.io",
+        nats.ErrorHandler(func(nc *nats.Conn, s *nats.Subscription, err error) {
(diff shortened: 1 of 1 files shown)


developing-with-nats/receiving/structure.md
```diff
@@ -7,7 +7,10 @@ For example, to receive JSON you could do:
 {% tabs %}
 {% tab title="Go" %}
 ```go
-nc, err := nats.Connect("demo.nats.io")
+nc, err := nats.Connect("demo.nats.io",
+        nats.ErrorHandler(func(nc *nats.Conn, s *nats.Subscription, err error) {
+            log.Printf("Async error in %q/%q: %v", s.Subject, s.Queue, err)
```
kozlovic: The ErrorHandler can be invoked also for error that do not involve the subscription, for instance a publish violation, etc.. so in other words, `s` may be `nil` here, which would then cause a panic.

philpennock: Oops, I only looked at testing that side when I was using `s` not `s.Subject` when I tried to make it prettier.  Sorry.  Fixed.

PR #28, opened 2020-01-03
Fix linux instructions

elucidsoft:
Linux instructions were incorrect.

### DIFF
--- a/nats-server/installation.md
+++ b/nats-server/installation.md
@@ -74,12 +74,12 @@ Download the zip file matching your systems architecture, and unzip. For this ex
 
 > unzip nats-server.zip -d nats-server
 Archive:  nats-server.zip
-   creating: nats-server-v2.0.0-darwin-amd64/
-  inflating: nats-server-v2.0.0-darwin-amd64/README.md
-  inflating: nats-server-v2.0.0-darwin-amd64/LICENSE
-  inflating: nats-server-v2.0.0darwin-amd64/nats-server
+   creating: nats-server-v2.0.0-linux-amd64/
+  inflating: nats-server-v2.0.0-linux-amd64/README.md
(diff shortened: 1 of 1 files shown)


nats-server/installation.md
```diff
@@ -74,12 +74,12 @@ Download the zip file matching your systems architecture, and unzip. For this ex
...
-  inflating: nats-server-v2.0.0-darwin-amd64/README.md
-  inflating: nats-server-v2.0.0-darwin-amd64/LICENSE
-  inflating: nats-server-v2.0.0darwin-amd64/nats-server
+   creating: nats-server-v2.0.0-linux-amd64/
+  inflating: nats-server-v2.0.0-linux-amd64/README.md
+  inflating: nats-server-v2.0.0-linux-amd64/LICENSE
+  inflating: nats-server-v2.0.0-linux-amd64/nats-server
 
-> cp nats-server-v2.0.0darwin-amd64/nats-server /usr/local/bin
+> sudo cp nats-server-v2.0.0-linux-amd64/nats-server /usr/bin
```
kozlovic: I get the `-linux-` change, but I think we should keep `/usr/local/bin`, no?

gcolliso: If that is correct then yes - does `local` change if you are using `sudo`?

gcolliso: @elucidsoft thoughts on the comments above?

elucidsoft: local is not as common as /usr/bin for Linux distributions.

kozlovic: I believe that user executables should preferably be copied to `/usr/local/bin` and OS executables are usually the ones in `/usr/bin`. But I am fine with the change. Thanks!

PR #39, opened 2020-01-24
Mention nats-box in the docs for nsc

wallyqs:


wallyqs:
related to https://github.com/nats-io/nsc/issues/249

PR #40, opened 2020-01-28
Set GKE command to actually create 3 nodes

rchenzheng:
Fix typo


wallyqs:
This is actually not a typo... but it does need more explanation, I think what might be happening on your setup is that all 3 nodes are being created in the same availability zone, instead of a kubelet being on different ones (a regional cluster).  For example from the tool docs https://cloud.google.com/kubernetes-engine/docs/how-to/creating-a-regional-cluster

```
To create a regional cluster with six nodes (three zones with two nodes each, specified by --num-nodes):

gcloud container clusters create my-regional-cluster \
    --num-nodes 2 \
    --region us-west1
```

So with the above command you would be getting a 6 nodes with 2 on each regions.

gcolliso:
@wallyqs @rchenzheng Could we close this PR? 
New info posted to doc by @wallyqs PR#51

rchenzheng:
Yes, it can be closed Thanks @wallyqs

PR #43, opened 2020-01-30
Document `allow-responses`.

aricart:
this documents how to configure the server for allow-responses (and a hint to how to do it with nsc).

### DIFF
--- a/nats-server/configuration/securing_nats/authorization.md
+++ b/nats-server/configuration/securing_nats/authorization.md
@@ -30,17 +30,17 @@ The `permission` map provides additional properties for configuring a `permissio
 
 ## Allow Responses Map
 
-The `allow_responses` option dynamically allows publishing to reply subjects, this works well for service responders.
+The `allow_responses` option dynamically allows publishing to reply subjects and works well for service responders.
 When set to `true`, excepting any `publish` permissions, implicitly all publish permissions are denied unless it is the reply subject in a request received by the client.
 
-The `allow_responses` map also allow you to configure a maximum number of responses and how long to allow the dynamic permission after a message is received.
+The `allow_responses` map also allows you to configure a maximum number of responses and how long to allow the dynamic permission after a message is received.
 
 | Property | Description |
 | :--- | :--- |
 |  `max` | The maximum number of response messages that can be returned to a request. |
 | `expires` | The duration of time to allow a response. Values such as `1s`, `1m`, `1h` (1 second, minute, hour) etc can be specified. |
 
-If `allow_responses` set to `true` rather than a specific configuration, it defaults to the equivalent of `{ max: 1 }`.
+If `allow_responses` is set to `true` rather than a specific configuration, it defaults to the equivalent of `{ max: 1 }`.
 
 **Important Note** When using `nsc` to configure your users, you can specify the `--allow-pub-response` and `--response-ttl` to control these settings.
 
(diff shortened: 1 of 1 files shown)


nats-server/configuration/securing_nats/authorization.md
```diff
@@ -26,6 +28,22 @@ The `permission` map provides additional properties for configuring a `permissio
 
 **Important Note** NATS Authorizations can be _allow lists_, _deny lists_, or both. It is important to not break request/reply patterns. In some cases \(as shown below\) you need to add rules as above with Alice and Bob for the `_INBOX.>` pattern. If an unauthorized client publishes or attempts to subscribe to a subject that has not been _allow listed_, the action fails and is logged at the server, and an error message is returned to the client.
 
+## Allow Responses Map
+
+The `allow_responses` option dynamically allows publishing to reply subjects, this works well for service responders.
+When set to `true`, excepting any `publish` permissions, implicitly all publish permissions are denied unless it is the reply subject in a request received by the client.
```
derekcollison: excepting any explicit 'allow' publish permissions...

nats-server/configuration/securing_nats/authorization.md
```diff
@@ -26,6 +28,22 @@ The `permission` map provides additional properties for configuring a `permissio
...
+
+The `allow_responses` map also allow you to configure a maximum number of responses and how long to allow the dynamic permission after a message is received.
```
derekcollison: Would start with. By default only one response is allowed, meaning the permission to publish to the reply subject defaults to only once. The allow_responses map does allow you to configure....

nats-server/configuration/securing_nats/authorization.md
```diff
@@ -26,6 +28,22 @@ The `permission` map provides additional properties for configuring a `permissio
...
+
+| Property | Description |
+| :--- | :--- |
+|  `max` | The maximum number of response messages that can be returned to a request. |
+| `expires` | The duration of time to allow a response. Values such as `1s`, `1m`, `1h` (1 second, minute, hour) etc can be specified. |
```
derekcollison: expires I think better described as, the amount of time the permission is valid.

nats-server/configuration/securing_nats/authorization.md
```diff
@@ -26,6 +28,22 @@ The `permission` map provides additional properties for configuring a `permissio
...
+
+If `allow_responses` set to `true` rather than a specific configuration, it defaults to the equivalent of `{ max: 1 }`.
```
derekcollison: Should add in "and no time limit"..

nats-server/configuration/securing_nats/authorization.md
```diff
@@ -95,3 +113,18 @@ authorization: {
...
+authorization: {
+	users: [
+		{ user: a, password: a },
+		{ user: b, password: b, permissions: {subscribe: "q", allow_responses: true } },
+		{ user: c, password: c, permissions: {subscribe: "q", allow_responses: { max: 5, expires: "1m" } } }
+	]
+}
+```
+
+User `a` has no restrictions. User `b` can listen on `q` for requests. If the received request has a reply subject it can reply at most one message. User `c` can also listen on `q` for requests, but is able to return at most 5 reply messages. The reply subject can be publish to for at most `1` minute.
```
derekcollison: User 'b' can subscribe/listen on subject 'q', and by default can only publish messages to reply subjects. All other subjects will be denied. ...

PR #44, opened 2020-02-01
Fix missing 'nats-io/' in a URL

philpennock:
Reported on Slack by https://github.com/rchenzheng


gcolliso:
lgtm - thanks @rchenzheng

PR #46, opened 2020-02-03
updating logging to include log_size_limit

gcolliso:
Adding log_size_limit, NATS Server v2.1.4

### DIFF
--- a/nats-server/configuration/logging.md
+++ b/nats-server/configuration/logging.md
@@ -64,12 +64,15 @@ All of these settings are available in the configuration file as well.
 debug:   false
 trace:   true
 logtime: false
+log_size_limit: 1GB
 log_file: "/tmp/nats-server.log"
 ```
 
-### Log Rotation with logrotate
+### Log Rotation
 
-NATS server does not provide tools to manage log files, but it does include mechanisms that make log rotation simple. We can use this mechanism with [logrotate](https://github.com/logrotate/logrotate); a simple standard Linux utility to rotate logs available on most distributions like Debian, Ubuntu, RedHat \(CentOS\), etc.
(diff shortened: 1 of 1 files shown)


gcolliso:
My first draft - I'm sure it will need help to be accurate and correct.

nats-server/configuration/logging.md
```diff
@@ -64,12 +64,15 @@ All of these settings are available in the configuration file as well.
 debug:   false
 trace:   true
 logtime: false
+log_size_limit: 1GB
```
kozlovic: Since I suggest we don't document the command line parameter (which is -log_size_limit), I would use here the `logfile_size_limit` name version which hints that it applies to logs that are of "file" nature.

nats-server/configuration/logging.md
```diff
@@ -64,12 +64,15 @@ All of these settings are available in the configuration file as well.
...
 log_file: "/tmp/nats-server.log"
 ```
 
-### Log Rotation with logrotate
+### Log Rotation
 
-NATS server does not provide tools to manage log files, but it does include mechanisms that make log rotation simple. We can use this mechanism with [logrotate](https://github.com/logrotate/logrotate); a simple standard Linux utility to rotate logs available on most distributions like Debian, Ubuntu, RedHat \(CentOS\), etc.
+Introduced in NATS Server v2.1.4, NATS allows for auto-rotation of log files when the size is greater than the configured limit. The backup files will have the same name as the original log file with the suffix .yyyy.mm.dd.hh.mm.ss.micros.
```
kozlovic: Should we hint to the name of the configuration param (`logfile_size_limit`) anywhere in this paragraph?

nats-server/configuration/logging.md
gcolliso: changed

nats-server/configuration/logging.md
gcolliso: added the param in the paragraph - that enough?

PR #52, opened 2020-02-11
Adding explanations/links and grouping of config options

matthiashanel:
also adding defaults
refreshed some sample output


gcolliso:
lgtm

PR #56, opened 2020-02-13
Fix wording and samples and add missing map properties

matthiashanel:
I fixed what was unclear to me or could be clearer.
Refreshed the examples.
added missing config properties. (specifically authorization)
Fixed the example to make it use one user.
(unfortunately, had to get rid of mkpasswd. that password does not work in an url. and having the unencrypted form in the same file defeats the point)

### DIFF
--- a/nats-server/configuration/clustering/README.md
+++ b/nats-server/configuration/clustering/README.md
@@ -21,22 +21,22 @@ Here is a simple cluster running on the same machine:
 
 ```bash
 # Server A - the 'seed server'
-> nats-server -p 4222 -cluster nats://0.0.0.0:5222
+> nats-server -p 4222 -cluster nats://localhost:4248
 
 # Server B
-> nats-server -p -1 -cluster nats://0.0.0.0:-1 -routes nats://localhost:5222
+> nats-server -p 5222 -cluster nats://localhost:5248 -routes nats://localhost:4248
 # Check the output of the server for the selected client and route ports.
 
 # Server C
-> nats-server -p -1 -cluster nats://0.0.0.0:-1 -routes nats://localhost:5222
+> nats-server -p 6222 -cluster nats://localhost:6248 -routes nats://localhost:4248
 # Check the output of the server for the selected client and route ports.
 ```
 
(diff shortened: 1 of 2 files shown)


nats-server/configuration/clustering/README.md
```diff
@@ -27,10 +32,12 @@ Here is a simple cluster running on the same machine:
...
+The _seed server_ has a client and cluster port specified. Through the use of `-1` as port, the other server are configured to auto-select unused ports for both clients and cluster connections. They then establish a route to the seed server. Because the clustering protocol gossips members of the cluster, all servers are able to discover other server in the cluster. When a server is discovered, the discovering server will automatically attempt to connect to it in order to form a _full mesh_. Typically only one instance of the server will run per machine, so you can reuse the client port \(4222\) and the cluster port \(5222\), and simply the route to the host/port of the seed server.
```
kozlovic: > Through the use of `-1` as port

This is not required and to be honest, except for tests where we want to minimize port conflicts if test packages were to be run in parallel (which we no longer do), I would not configure production servers with "random" ports. That may be just me though :-).

nats-server/configuration/clustering/README.md
```diff
@@ -27,10 +32,12 @@ Here is a simple cluster running on the same machine:
...
 
 Similarly, clients connecting to any server in the cluster will discover other servers in the cluster. If the connection to the server is interrupted, the client will attempt to connect to all other known servers.
 
+There is no explicit configuration for _seed server_. They simply serve as the starting point for server discovery by other members of the cluster as well as clients. As such these are the server that clients have in their list of connect urls and cluster members have in their list of routes. They reduce configuration as not every server needs to be in these lists. But the ability for other server and clients to successfully connect depends on _seed server_ running.
```
kozlovic: missing "s" in "As such these are the server that.."

matthiashanel: fixed

nats-server/configuration/clustering/README.md
matthiashanel: Agreed, I removed -1 and it's explanation.
Fixed the example that goes with this.
Called out routes option when using multiple seed server

gcolliso:
lgtm - thanks @matthiashanel

PR #57, opened 2020-02-14
Rewording and clarifying added missing properties

matthiashanel:
Also described message flow in interest propagation section

### DIFF
--- a/nats-server/configuration/gateways/README.md
+++ b/nats-server/configuration/gateways/README.md
@@ -21,11 +21,11 @@ Gateways exist to:
 * reduce the number of connections required between servers 
 * optimize the interest graph propagation
 
-If gateways are to be used in a cluster, **all** servers of this cluster need to have a gateway configuration with the **same name**. Furthermore, every gateway node needs to be able to **connect to every** other gateway node and vice versa. Everything else is considered a misconfiguration.
+If gateways are to be used in a cluster, **all** server of this cluster need to have a gateway configuration with the **same name**. Furthermore, every gateway node needs to be able to **connect to any** other gateway node and vice versa. Everything else is considered a misconfiguration.
 
 ## Gateway Connections
 
-A nats-server in a gateway role will specify a port where it will accept gateway connections. If the configuration specifies other _external_ `gateways`, the gateway will create one outbound gateway connection for each gateway in it's configuration. It will also gossip other gateways it knows or discovered. Fewer _external_ `gateways` mean less configuration. Ye, the ability to discover more gateways and gateway nodes depends on these servers running. This is similar to _seed server_ in cluster. It is recommended to have all _seed server_ of a cluster listed in the `gateways` section.
+A nats-server in a gateway role will specify a port where it will accept gateway connections. If the configuration specifies other _external_ `gateways`, the gateway will create one outbound gateway connection for each gateway in it's configuration. It will also gossip other gateways it knows or discovers. Fewer _external_ `gateways` mean less configuration. Yet, the ability to discover more gateways and gateway nodes depends on these servers running. This is similar to _seed server_ in cluster. It is recommended to have all _seed server_ of a cluster listed in the `gateways` section.
 
 If the local cluster has three gateway nodes, this means there will be three outbound connections to each external gateway.
 
@@ -61,8 +61,8 @@ Messages from clients directly connected to a gateway node will be sent along ou
 * Interest-only Mode
 * Queue Subscriptions
 
-Local interest permitting, the receiving gateway node sends the messages directly to it's subscribing clients as well as server within the cluster that have subscribing clients.
-
+Local interest permitting, the receiving gateway node sends the messages directly to it's subscribing clients as well as server within the cluster.
+ 
 ### Optimistic Mode
 
 When a publisher in _A_ publishes "foo", the _A_ gateway will check if cluster _B_ has registered _no_ interest in "foo". If not, it forwards "foo" to _B_. If upon receiving "foo", _B_ has no subscribers on "foo", _B_ will send a gateway protocol message to _A_ expressing that it has no interest on "foo", preventing future messages on "foo" from being forwarded.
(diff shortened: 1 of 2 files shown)


nats-server/configuration/gateways/README.md
```diff
@@ -2,26 +2,30 @@
...
-* gateways gossip gateway members and remote discovered gateways
+* gateways gossip gateway nodes and remote discovered gateways
 
 Unlike clusters, gateways:
 
-* don't form a full mesh
+* have names, specifying the cluster they are in
+* don't form a full mesh between gateway nodes, but form a full mesh between cluster instead
 * are bound by uni-directional connections
+* don't gossip gateway nodes to clients
```
kozlovic: I think I know what you mean here, but to be clear, nodes from a cluster are not gossiped to clients connected to another cluster. That is, if cluster A has 3 servers A1, A2 and A3, and cluster B has 2 servers B1 and B2, a client connecting to A1 will be notified of A2 and A3, but no server from B. Same for cluster B, a client connecting to B1 will be notified of B2 but no server from cluster A.

nats-server/configuration/gateways/README.md
```diff
@@ -2,26 +2,30 @@
...
 
 Gateways exist to:
 
 * reduce the number of connections required between servers 
 * optimize the interest graph propagation
 
+If gateways are to be used in a cluster, **all** server of this cluster need to have a gateway configuration with the **same name**. Furthermore, every gateway node needs to be able to **connect to every** other gateway node and vice versa. Everything else is considered a miss configuration.
```
kozlovic: > be able to **connect to every** other gateway node

That could be interpreted as a node in cluster A needs to be able to connect to all nodes in cluster B, as in creating a connection to all nodes. As you know, this is not the case. Each node needs to be able to "connect to any" other gateway node.

> miss configuration

I think "misconfiguration" instead?

nats-server/configuration/gateways/README.md
```diff
@@ -2,26 +2,30 @@
...
+
 ## Gateway Connections
 
-A nats-server in a gateway role will specify a port where it will accept gateways connections. If the configuration specifies other _external_ `gateways`, the gateway will create one outbound gateway connection for each gateway in its configuration. It will also gossip other gateways it knows or discovered.
+A nats-server in a gateway role will specify a port where it will accept gateway connections. If the configuration specifies other _external_ `gateways`, the gateway will create one outbound gateway connection for each gateway in its configuration. It will also gossip other gateways it knows or discovered. Fewer _external_ `gateways` mean less configuration. Ye, the ability to discover more gateways and gateway nodes depends on these server running. This is similar to _seed server_ in cluster. It is recommended to have all _seed server_ of a cluster listed in the `gateways` section.
```
kozlovic: > other gateways it knows or discovered. 

or "discovers" to be in the same tense?

> Ye, the ability to discover

Yet?

nats-server/configuration/gateways/README.md
```diff
@@ -70,7 +77,7 @@ When a gateway on _A_ sends many messages on various subjects for which _B_ has
 
 When a queue subscriber creates a new subscription, the gateway propagates the subscription interest to other gateways. The subscription interest is only propagated _once_ per _Account_ and subject. When the last queue subscriber is gone, the cluster interest is removed.
 
-Queue subscriptions work on _Interest-only Mode_ to honor NATS' queue semantics across the _Super Cluster_. For each queue group, a message is only delivered to a single queue subscriber. Only when a local queue group member is not found, is a message forwarded to a different interested cluster; gateways will always try to serve local queue subscribers first and only failover when a local queue subscriber is not found.
+Queue subscriptions work on _Interest-only Mode_ to honor NATS' queue semantics across the _Super Cluster_. For each queue group, a message is only delivered to a single queue subscriber. Only when a local queue group member is not found, is a message forwarded to a different interested cluster; gateways will always try to serve local queue subscribers first and only failover when a local queue subscriber is not found. The interested cluster is picked based on lowest RTT.
```
kozlovic: > Only when a local queue group member is not found, is a message forwarded to a different interested cluster; gateways will ...

Would it read better like this?
"If the same queue group exists in multiple clusters, the server will pick one member from the queue group in its cluster, only sending to a different cluster if there is no interest in its cluster. In other words, a server will always try.."

nats-server/configuration/gateways/README.md
```diff
@@ -48,14 +52,17 @@ The number of connections required to join clusters using clustering vs. gateway
...
 
-Gateways propagate interest using three different mechanisms:
+## Interest Propagation
+Messages from clients directly connected to a gateway node will be sent along outgoing gateway connections according to the following three interest propagation mechanisms:
 
 * Optimistic Mode
 * Interest-only Mode
 * Queue Subscriptions
 
+Local interest permitting, the receiving gateway node sends the messages directly to it's subscribing clients as well as server within the cluster that have subscribing clients.
```
gcolliso: ...to it's subscribing clients as well as <any><all><a> server within the cluster<s> that have ...

nats-server/configuration/gateways/README.md
matthiashanel: This in the section where gateways are contrasted with cluster
Unlike clusters, gateways:
...
* don't gossip gateway nodes to clients

That is essentially what you are saying.

nats-server/configuration/gateways/README.md
matthiashanel: changed to any and misconfiguration

nats-server/configuration/gateways/README.md
matthiashanel: fixed

nats-server/configuration/gateways/README.md
matthiashanel: agreed, also changed the last sentence to:
. The server will pick the cluster with the lowest RTT.

nats-server/configuration/gateways/README.md
matthiashanel: removed

nats-server/configuration/gateways/README.md
```diff
@@ -70,7 +77,7 @@ When a gateway on _A_ sends many messages on various subjects for which _B_ has
...
 
-Queue subscriptions work on _Interest-only Mode_ to honor NATS' queue semantics across the _Super Cluster_. For each queue group, a message is only delivered to a single queue subscriber. Only when a local queue group member is not found, is a message forwarded to a different interested cluster; gateways will always try to serve local queue subscribers first and only failover when a local queue subscriber is not found.
+Queue subscriptions work on _Interest-only Mode_ to honor NATS' queue semantics across the _Super Cluster_. For each queue group, a message is only delivered to a single queue subscriber. If the same queue group exists in multiple clusters, the server will pick one member from the queue group in its cluster, only sending to a different cluster if there is no interest in its cluster. In other words, a server will always try to serve local queue subscribers first and only failover when a local queue subscriber is not found. The server will pick the cluster with the lowest RTT.
```
kozlovic: >  only sending to a different cluster if there is no interest in its cluster.

After re-reading it, `in its cluster` seem redundant. So simply `if there is no local interest`. It's just that I hope user do not interpret "local" with server-only. In the context of gateways and subscription interest, here "local" means all nodes in a given cluster.

nats-server/configuration/gateways/README.md
```diff
@@ -2,26 +2,30 @@
...
+
 ## Gateway Connections
 
-A nats-server in a gateway role will specify a port where it will accept gateways connections. If the configuration specifies other _external_ `gateways`, the gateway will create one outbound gateway connection for each gateway in its configuration. It will also gossip other gateways it knows or discovered.
+A nats-server in a gateway role will specify a port where it will accept gateway connections. If the configuration specifies other _external_ `gateways`, the gateway will create one outbound gateway connection for each gateway in it's configuration. It will also gossip other gateways it knows or discovers. Fewer _external_ `gateways` mean less configuration. Yet, the ability to discover more gateways and gateway nodes depends on these servers running. This is similar to _seed server_ in cluster. It is recommended to have all _seed server_ of a cluster listed in the `gateways` section.
```
kozlovic: > the gateway will create one outbound gateway connection for each gateway in it's configuration.

@gcolliso I think the original `its` was correct. The meaning here is "possessive", so "its" is the correct word.

PR #63, opened 2020-03-04
Clustering: add note about horizontal scalability

kozlovic:
Since NATS Streaming implementation is such as the leader is the one handling all communication with clients, adding a note that the cluster cannot be horizontally scaled, and that its size should be limited to 3 to 5 nodes.


kozlovic:
This relates to https://github.com/nats-io/nats-streaming-server/issues/999.

PR #64, opened 2020-03-11
tls chagnes

matthiashanel:
Add a section on cert generation. (got Phil's input and ok)

Move from test certificates to generated ones. 
(Will do that for the client in a separate step. Java complicates things)

Mention verify, verify_and_map and monitoring.
Mention order in which user name is found in certs.

### DIFF
--- a/nats-server/configuration/securing_nats/tls.md
+++ b/nats-server/configuration/securing_nats/tls.md
@@ -88,7 +88,7 @@ While this works for server and libraries from the NATS eco system, you will exp
 Another option is to configure your system's trust store to include self signed certificate(s).
 Which trust store needs to be configured depends on what you are testing.
 * This may be your OS for server and certain clients.
-* The runtime environment for other clients like Java or Python.
+* The runtime environment for other clients like Java, Python or Node.js.
 * Your browser for monitoring endpoints and websockets.
 
 Please check your system's documentation on how to trust a particular self signed certificate.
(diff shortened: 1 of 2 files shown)


nats-server/configuration/securing_nats/tls.md
```diff
@@ -59,3 +59,90 @@ tls: {
...
+This zero config tool generates and installs the CA into your **local** system trust store(s) and makes providing SAN straight forward. 
+Here is an example:
+
+Generate a CA and output the location of the root CA cert file `rootCA.pem`. 
+Next generate a certificate, valid for use by `localhost` and the IP `::1`(`-cert-file` and `-key-file` overwrite default file names).
+Then start a nats server using the generated certificate.
+
+```bash
+mkcert -install
+mkcert -CAROOT
```
kozlovic: If I am not mistaken, this simply displays the location of where the CA root cert will be stored. Unless you have the example where you pass `--tls_cacert` to the `nats-server` exe, then I am not sure you need to have that.

matthiashanel: Verify does just that... I also mentioned that ca_file defaults to the system store.  
I will move mkcert-CAROOT further down and turn that into a separate step.

matthiashanel: Moved it into a separate section saying:
Examples in this document make use of the certificates generated so far.
To simplify examples using the CA certificate, copy `rootCA.pem` into the same folder where the certificates were generated. To obtain the CA certificate's location use this command:

gcolliso:
lgtm

PR #71, opened 2020-03-18
Add example of getting started with nsc + nats account-server

wallyqs:


nats-tools/nas/nas_conf.md
```diff
@@ -156,3 +156,56 @@ Let's take a look at the configuration options:
...
+## Example Setup
+
+Below you can find an example of setting up a local environment 
+using nsc to create 4 accounts, one of them a system account, 
+and setting the account server URL to be used for the operator.
+
+```sh
+$ export NKEYS_PATH=$(pwd)/nsc/nkeys
+$ export NSC_HOME=$(pwd)/nsc/accounts
+$ curl -sSL https://nats-io.github.io/k8s/setup/nsc-setup.sh | sh
```
matthiashanel: To me this does not quite match the expectations set in the description.
This example downloads a script to generate 4 accounts using nsc.
This script does most of what is in the description.

Maybe write something like this?:

Provided a setup with 4 accounts, one of them a system account, this example shows how to set up the account server by:
* adding the account server to the operator
* configuring the account server
* push the accounts to the account server
* configure a `nats-server` to make use of the account server
* test the setup

wallyqs: thanks, that's much better 👍

wallyqs: ok made some updates

PR #88, opened 2020-04-14
Mentioning response_type in accounts.

matthiashanel:
So far I see no need to mention chunked yet.
As far as I can tell in the code, it's singleton or not singleton.


matthiashanel:
noticed I had put it in the wrong section. Is an export option.

@kozlovic , while the app may not have any way to know, the requestor won't be able to receive more than one message... So we should document a way to turn it off.
nsc shows these options too. And documenting them here fits with the story of jwt are accounts but distributed.

nsc exposes chunked too, but right now I really don't know what to write there.

kozlovic:
@matthiashanel I did not mean to not document it, on the contrary, I was saying that we could add "chunked" as a possible value. But then, like I said, not sure how user can use that information anyway. So again, no question that "singleton", "stream" need to be documented, just wondering if we should list "chunked" as a choice too.

gcolliso:
@matthiashanel What would you like to do about this PR?

matthiashanel:
@gcolliso I'd merge it. we can add chunked later or once it makes a difference.

gcolliso:
lgtm - thanks @matthiashanel

PR #89, opened 2020-04-21
clarifications for mkcert -client

matthiashanel:


matthiashanel:
these changes not ok?

PR #96, opened 2020-05-06
Add note on using L4 load balancers in K8S

wallyqs:


gcolliso:
@wallyqs Since this is a new document, where do you want it added in the outline?

wallyqs:
thanks, I think at the bottom from here would be good: https://github.com/nats-io/nats.docs/blob/de933e912d729f968b365734baae5df8f0f9a731/SUMMARY.md#nats-on-kubernetes

gcolliso:
lgtm

PR #98, opened 2020-05-13
Documenting monitoring endpoint as system service

matthiashanel:
please merge only after ok+server release.

I have neglected to mention what possible options are.
On my list of todos is creating json schema for input as well as output.
My intention is to list the schemas in the table added.


gcolliso:
lgtm

PR #107, opened 2020-05-21
Create ngs-docker-python.md

wallyqs:


wallyqs:
Let's add it inside of the nats docker section here for the summary
Screen Shot 2020-05-21 at 2 23 02 PM

PR #124, opened 2020-07-02
Update tls.md

vadim-shadrin:


gcolliso:
@wallyqs is this something you could review?

gcolliso:
@wallyqs trying to clean up .docs - could you review this change and let me know if it's ok to merge or update where necessary?

wallyqs:
need to update to use thale nats.py client under nats-io ideally, maybe close for now

PR #132, opened 2020-07-27
no persistence in plain nats

matthiashanel:
adding "It does not store messages on disk. "


nats-concepts/intro.md
```diff
@@ -6,5 +6,5 @@ NATS makes it easy for programs to communicate across different environments, la
...
 
-NATS core offers an **at most once** quality of service. If a subscriber is not listening on the subject \(no subject match\), or is not active when the message is sent, the message is not received. This is the same level of guarantee that TCP/IP provides. By default, NATS is a fire-and-forget messaging system. If you need higher levels of service, you can use [NATS Streaming](../nats-streaming-concepts/intro.md) or build additional reliability into your client applications with proven and scalable reference designs such as [acks](acks.md) and [sequence numbers](seq_num.md).
+NATS core offers an **at most once** quality of service. If a subscriber is not listening on the subject \(no subject match\), or is not active when the message is sent, the message is not received. This is the same level of guarantee that TCP/IP provides. NATS is a fire-and-forget messaging system. It does not store messages on disk. If you need higher levels of service, you can use [NATS Streaming](../nats-streaming-concepts/intro.md) or build additional reliability into your client applications with proven and scalable reference designs such as [acks](acks.md) and [sequence numbers](seq_num.md).
```
ColinSullivan1, line 9: Thanks!

How about, "Core NATS will only hold messages in memory and will never write messages directly to disk."  We'll want to be careful with verbiage - the OS _may_ write memory containing messages to disk based on swapping - that's out of our control.

matthiashanel: agreed

matthiashanel:
@ColinSullivan1 changed it to: `Core NATS is a fire-and-forget messaging system, will only hold messages in memory and will never write messages directly to disk.`

gcolliso:
lgtm

PR #135, opened 2020-07-28
Update STAN ft docs to use static NATS routes

wallyqs:


lundbird:
@wallyqs What was the reason this change was made?

wallyqs:
@lundbird when not using static routes, there could be some scenarios when there are partitions where stan-1 connects to self and does not retry and stan-2 and stan-3 to each other due to the way that the headless dns service discovery works.  There is a recent change merged recently that may help improve this by detecting when resolving self route when the cluster is starting: https://github.com/nats-io/nats-server/pull/1590

PR #136, opened 2020-07-29
Update swarm.md

lucj:
Use synadia:nats-box instead of golang:latest (no stan.go in that one)


jnmoyne (approved):
Cleaning up old PR, LGTM (it's all legacy now anyways)

PR #150, opened 2020-09-04
fix nc.PublishRequest paramer

KingDacker:
if err := nc.PublishRequest("time", uniqueReplyTo, nil); err != nil {
    log.Fatal(err)
}
========>
if err := nc.PublishRequest(uniqueReplyTo, "time", nil); err != nil {
    log.Fatal(err)
}

### DIFF
--- a/developing-with-nats/sending/replyto.md
+++ b/developing-with-nats/sending/replyto.md
@@ -22,7 +22,7 @@ if err != nil {
 
 // Send the request.
 // If processing is synchronous, use Request() which returns the response message.
-if err := nc.PublishRequest("time", uniqueReplyTo, nil); err != nil {
+if err := nc.PublishRequest(uniqueReplyTo, "time", nil); err != nil {
     log.Fatal(err)
 }
 


gcolliso:
@kozlovic could you confirm the proposed change, please? Thank you!

developing-with-nats/sending/replyto.md
```diff
@@ -22,7 +22,7 @@ if err != nil {
 
 // Send the request.
 // If processing is synchronous, use Request() which returns the response message.
-if err := nc.PublishRequest("time", uniqueReplyTo, nil); err != nil {
+if err := nc.PublishRequest(uniqueReplyTo, "time", nil); err != nil {
```
kozlovic, line 25: This is wrong. The original code is correct: you send the request to the public subject "time" and set the reply subject to `uniqueReplyTo` (second parameter). This will set the published message subject to "time" and reply subject to `uniqueReplyTo`

kozlovic:
Closing this PR since the change is incorrect and original code example is correct.

PR #152, opened 2020-09-06
Fix typo

variadico:
We typed "Authenticatin" instead of "Authentication"


gcolliso (commented):
lgtm - thanks @variadico

PR #157, opened 2020-09-27
Update shard config option description.

dontlaugh:
shard -> shared

### DIFF
--- a/nats-tools/nas/nas_conf.md
+++ b/nats-tools/nas/nas_conf.md
@@ -119,7 +119,7 @@ Let's take a look at the configuration options:
 | `dir` | Configures a directory as a store. |
 | `nsc` | Configures an nsc read-only store. The value should be the path to an operator _directory_. Option is mutually exclusive with `dir`. |
 | `readonly` | If `true`, the store will not accept POST requests. Note that to receive requests, the store must also have `operatorjwtpath` specified as a root option. |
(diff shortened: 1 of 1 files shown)


dontlaugh:
Hey folks, I actually don't know if this config param is right, but it _looks_ like it should be "shared". Is that the case?

(I haven't deployed a jwt server yet, so I don't have an opportunity to find out for myself)

matthiashanel:
Hello @dontlaugh , it is shard from sharding. This option is to avoid having too many files in one directory by means of sharding them. That is putting them in subdirectories named after the last digits(s) in the account name. Does this make sense?

matthiashanel (changes requested):
it is shard from sharding and does not relate to shared.

dontlaugh:
Ah, I see now. Thanks :) So perhaps something like "if true, JWTs will be stored in multiple directories" or "JWTs will be automatically stored in multiple directories"?

matthiashanel:
Yeah I like that. This will also fix the faulty description. I would maybe include smth like this: stored in sub directories of the store directory.

dontlaugh:
@matthiashanel take another look?

matthiashanel (approved):
LGTM

matthiashanel:
@dontlaugh just did LGTM. Thanks for your contribution!!!

gcolliso:
lgtm - thanks!

PR #160, opened 2020-10-26
Update no_advertise comment

kozlovic:
Resolves https://github.com/nats-io/nats-server/issues/1647


nats-server/configuration/clustering/cluster_config.md
```diff
@@ -9,7 +9,7 @@ The `cluster` configuration map has the following configuration options:
 | `listen` | Combines `host` and `port` as `<host>:<port>`. |
 | `tls` | A [`tls` configuration map](../securing_nats/tls.md) for securing the clustering connection. `verify` is always enabled and `cert_file` is used for client and server. [See](../securing_nats/tls.md#wrong-key-usage) for certificate pitfalls. |
 | `advertise` | Hostport `<host>:<port>` to advertise how this server can be contacted by other cluster members. This is useful in setups with NAT. |
-| `no_advertise` | When set to `true`, the server will not communicate the URLs in which it accepts client connections to the other servers in the cluster, but also the other servers' URLs to its own clients. |
+| `no_advertise` | When set to `true`, the server will not communicate the URLs it accepts from client connections to the other servers in the cluster, as well as the other servers' URLs to the server's own clients. |
```
kozlovic, line 12: I wonder if this does not change the meaning of my original text (but it may just be me failing to understand english). The way you phrase it sounds to be that the server "will not communicate the URLs it accepts from client connections" as if clients were sending it URLs. 
My intention was to explain that the server will not communicate to other servers the URLs it uses to accept client connections.

kozlovic: I am having difficulties explaining this: let me make it long so that you can summarize in english:
- the server is configured to accept client connections using either a host name or IP and a port.
- when it is a hostname or the "any" IP (0.0.0.0), the server will resolve that to actual IPs, there may be more than one.
- this list is normally sent to other servers so that those servers can tell clients that connect to them about these URLs of that server. This allow client libraries to "learn" about other servers they could reconnect to should their connection to their server be lost.

This property disables all that from a server perspective. Of course, servers in the same cluster should have the same configuration, but they don't have to. So if no_advertise=true for a given server, I wanted to express that this server will not send its "client URLs" to the other servers and also won't tell its client about the other servers "client URLs".
Does that make sense?

gcolliso: Can we just use your last statement then? Something like...
When set to 'true', the server will not send or gossip its client URLs to other servers in the cluster and will not tell its client about the other servers' client URLs.

kozlovic: Sure, it's just that I was concerned that users may not know what "client URL" mean, but if you think that is pretty clear, then I am ok with that.

gcolliso: I think it is ok - we can always revisit if we get questions.

kozlovic: Sounds good!

PR #164, opened 2020-10-30
fix links and wording

matthiashanel:
noticed broken links


gcolliso:
lgtm - thank @matthiashanel

PR #172, opened 2020-11-18
clarify insecure

matthiashanel:
This came up today...   

This is the go documentation of the value that this configuration sets

	// InsecureSkipVerify controls whether a client verifies the
	// server's certificate chain and host name.
	// If InsecureSkipVerify is true, TLS accepts any certificate
	// presented by the server and any host name in that certificate.
	// In this mode, TLS is susceptible to man-in-the-middle attacks.
	// This should be used only for testing.
	InsecureSkipVerify bool


kozlovic (approved):
LGTM. Note that the server will print a warning when this is set. It applies to Gateway, Leafnode and Route for outgoing connections, as you state here. Example for gateways:
```
TLS certificate chain and hostname of solicited gateways will not be verified. DO NOT USE IN PRODUCTION!
```

gcolliso:
lgtm - thank you @matthiashanel & @kozlovic

PR #181, opened 2020-12-14
Fix link to client-side slow consumer detection

watfordgnf:
Noticed the link to client-side slow consumers detection was missing.


gcolliso (commented):
lgtm - thank you @watfordgnf

PR #184, opened 2020-12-22
Maybe it was inverted Server and streaming server?

ste4net:


kozlovic (commented):
No, the original stance is correct. The NATS Streaming server is not a server per-se. It is a program that uses NATS library to connect to a NATS Server (embedded or not). So streaming clients are not connecting to a NATS Streaming "server", but simply connect to a NATS Server. Streaming server and clients are NATS clients that communicate using NATS, but creating NATS subscriptions and publishing NATS messages.

kozlovic:
I would close the PR but will let @ste4net read my comment and see if some more need to be added to the documentation to make that more clear.

gcolliso:
@ste4net Did you need additional clarification for this issue? If not, I will close it.

ste4net:
No thanks. Sorry for missing reply


Il mar 9 feb 2021, 18:14 Ginger Collison <notifications@github.com> ha
scritto:

PR #188, opened 2020-12-31
Add explicit account server URL to `nsc push` step

bruth:


gcolliso:
@aricart or @matthiashanel would you be able to review, please?

nats-tools/nas/dir_store.md
```diff
@@ -63,7 +63,7 @@ Success! - added account "B"
 With the account and a couple of users in place, let's push all the accounts to the nats-account-server:
 
 ```text
-> nsc push -A
+> nsc push -u http://localhost:9090/jwt/v1/ -A
```
aricart, line 66: This is only required if the operator doesn't set an account server URL. 
To simplify the workflow, you can edit the operator:
```
❯ nsc edit operator -u http://localhost:9090/jwt/v1
[ OK ] set account jwt server url to "http://localhost:9090/jwt/v1"
[ OK ] edited operator "AAA"

```

aricart (changes requested):
LGTM but the reason why the url is required is because the operator is not correctly configured. An account server URL should be set. We should add the bit of the comment, and then show that even if not set, specifying the URL will work.

nats-tools/nas/dir_store.md
matthiashanel: I concur with @aricart

bruth: Thanks, understood. It was not clear when reading through the docs that this was necessary or preferred over the `push -u` option. That is why I wanted to raise it here. Of course feel free to update the docs however is appropriate.

bruth:
@aricart Updated the change to include the note.

matthiashanel (approved):
LGTM

aricart (approved):
LGTM

bruth:
Thanks all, changes look good to me as well.

gcolliso (commented):
lgtm - thanks all!

PR #189, opened 2021-01-01
Fix a poorly-formatted link

skyzyx:
* Fix a poorly-formatted link.
* Used the [most usable link pattern](https://www.nngroup.com/articles/writing-links/) originally documented in 1996.


gcolliso (commented):
Thanks @skyzyx

Issue #183, opened 2020-12-15
Unable to install mkpasswd like in documentation


moleksyuk:
Hi guys,

I'm trying to install `mkpasswd` from [this documentation](https://docs.nats.io/nats-tools/mkpasswd#installing-mkpasswd).

**Issue No.1** - Wrong symbol in command
Copy button copies content `> go get github.com/nats-io/nats-server/util/mkpasswd` with starting `>`. It makes command invalid and getting error: `-bash: go: Is a directory`

**Issue No.2** - Unable to install mkpasswd

**Steps to reproduce:**
1. Install **go**: `sudo snap install go --classic`
2. Install **mkpasswd**: `go get github.com/nats-io/nats-server/util/mkpasswd`

Actual: getting error
```
~$ go get github.com/nats-io/nats-server/util/mkpasswd
package github.com/nats-io/nats-server/util/mkpasswd: build constraints exclude all Go files in /home/ubuntu/go/src/github.com/nats-io/nats-server/util/mkpasswd
```

Expected: mkpasswd is installed.

OS: 
```Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.1 LTS
Release:	20.04
Codename:	focal
```
Go: 
`go version go1.15.6 linux/amd64`

Go env:
```
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/ubuntu/.cache/go-build"
GOENV="/home/ubuntu/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/ubuntu/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/ubuntu/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/snap/go/6745"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/snap/go/6745/pkg/tool/linux_amd64"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
[...]

PR #201, opened 2021-01-25
Remove shell prompt character

variadico:
Currently, our copy-able snippets have a shell prompt. When someone clicks on
the copy button, the shell prompt gets copied along with the command we want
copied. This removes the shell prompt so that users don't have to edit that out
themselves when they paste one of our commands.

Resolves https://github.com/nats-io/nats.docs/issues/183


variadico:
I'm not super familiar with how our docs work. Am I correct to assume that we update the markdown and then somehow generate the HTML?

ColinSullivan1 (approved):
LGTM.

gcolliso (commented):
lgtm

PR #209, opened 2021-01-30
nc is closed by the defer statement.

tmcnicol:


gcolliso:
@kozlovic could you review just to be sure this is accurate? Thank you!

developing-with-nats/events/README.md
```diff
@@ -25,8 +25,6 @@ getStatusTxt := func(nc *nats.Conn) string {
 }
 log.Printf("The connection is %v\n", getStatusTxt(nc))
 
-nc.Close()
```
kozlovic, line 28: The connection is closed by defer statement, but we explicitly close here to show the difference of the status. See how we print the status before and after the close to show the difference.

kozlovic:
It is true that there is a defer, but we want an explicit close here to show the status before and after the close.

developing-with-nats/events/README.md
tmcnicol: yeah that makes sense

PR #221, opened 2021-02-09
[removed] -nsc option and config from account server doc

matthiashanel:

we removed this from nats-account-server


aricart (approved):
LGTM

ColinSullivan1 (approved):
LGTM... When was the flag removed?  Just thinking about first deprecating this, then removing later, but if it's been gone for awhile, no need.

matthiashanel:
> LGTM... When was the flag removed? Just thinking about first deprecating this, then removing later, but if it's been gone for awhile, no need.

I removed that 6 Months ago,  after consulting with @aricart as part of the jwtv2 work.

PR #222, opened 2021-02-10
Clarify Example Gateway Connection Logic

bhass1:
The current description reads like each external gateway, B1 & B2, should have three connections from the local cluster because the local cluster has three gateway nodes. However, my understanding is that the external gateway _cluster_ will have three connections from the local cluster. 
Added cluster at the end of sentence and clarified there will be  three outbound connections from the local cluster.


gcolliso (commented):
lgtm - thanks @bhass1

PR #234, opened 2021-03-18
pointed at the nsc 2.2.0 release notes, as the new version of the tooling is not compatible with older servers.

aricart:


gcolliso:
lgtm - thanks @aricart

PR #245, opened 2021-04-19
Fix typo in nats_api_reference.md

jdrews:
I believe this should say "end" not "and"


gcolliso (commented):
lgtm - thanks @jdrews

PR #258, opened 2021-05-11
Document that config file is UTF-8

philpennock:


gcolliso (commented):
lgtm - thanks @philpennock

PR #259, opened 2021-05-12
Add note to same_origin and allowed_origins

kozlovic:
Clarify that these 2 options are used only for clients that present an Origin header, likely from web browsers.


gcolliso (commented):
lgtm - thanks @kozlovic

PR #260, opened 2021-05-13
Move section header out of paragraph text

davidmorley:
The text "Scalable, Future-Proof Deployments" appears to be a section header, but was included in the preceding paragraph.


gcolliso (commented):
lgtm - thanks @davidmorley

PR #261, opened 2021-05-13
Add section on naming streams, consumers, and accounts

ColinSullivan1:
Added a section on naming streams, consumers, and accounts.

Signed-off-by: Colin Sullivan <colin@synadia.com>


jetstream/administration/naming.md
```diff
@@ -0,0 +1,19 @@
...
+accounts, names must adhere to subject naming rules as well as being
+friendly to the file system.
+
+We recommend the following guideline for stream, consumer, and account names:
+
+* Alphanumeric values are recommended.
+* Spaces, tabs, period \(`.`\), greater than \(`>`\) or asterix \(`*`\) are prohibited.
+* Limit names to 255 characters or less.
+* Do not use reserved file names like NUL, LPT1, etc.
+* Be aware that some file systems are case insentitive so do not
```
gcolliso, line 15: Wouldn't it be file systems that are case sensitive that would be the issue?

ColinSullivan1: Actually, the issue is that it's case insensitive, but I'll look to clarify.  There is another spot I need to update as well.

gcolliso:
@ColinSullivan1 is this g2g from your side?

jetstream/administration/naming.md
ColinSullivan1: I've updated this PR.  On case insensitive systems, one could create a stream "Foo" and "foo" and directories would collide on disk.

ColinSullivan1:
Yep, g2g here!

gcolliso:
lgtm - thanks @ColinSullivan1

PR #265, opened 2021-06-08
Improve explanation of MQTT authentication in operator mode

mullerch:
- Indicate that the seed is not used
- Indicate that the JWT is used for authentication
- Indicate how to set the Bearer boolean with nsc

### DIFF
--- a/nats-server/configuration/mqtt/mqtt_config.md
+++ b/nats-server/configuration/mqtt/mqtt_config.md
@@ -104,9 +104,11 @@ mqtt {
 
 ### Operator mode
 
-In operator mode, all users need to provide a JWT in order to connect. For MQTT clients, it means that you need to pass the JWT token as the MQTT password and use any username since MQTT protocol requires a username to be set if a password is set.
+In operator mode, all users need to provide a JWT in order to connect. In the standard authentication procedure of this mode, NATS clients are required to sign a `nonce` sent by the server using their private key \(see [JWTs and Privacy](../securing_nats/jwt/#jwts-and-privacy)\). MQTT clients cannot do that, therefore, the JWT is used for authentication, removing the need of the seed. It means that you need to pass the JWT token as the MQTT password and use any username (except empty, since MQTT protocol requires a username to be set if a password is set). The JWT has to have the `Bearer` boolean set to true, which can be done with nsc:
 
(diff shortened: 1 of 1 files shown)


gcolliso:
@kozlovic could you review, please?

nats-server/configuration/mqtt/mqtt_config.md
```diff
@@ -104,9 +104,11 @@ mqtt {
 
 ### Operator mode
 
-In operator mode, all users need to provide a JWT in order to connect. For MQTT clients, it means that you need to pass the JWT token as the MQTT password and use any username since MQTT protocol requires a username to be set if a password is set.
+In operator mode, all users need to provide a JWT in order to connect. In the standard authentication procedure of this mode, NATS clients are required to sign a `nonce` sent by the server using their private key \(see [JWTs and Privacy](../securing_nats/jwt/#jwts-and-privacy)\). MQTT clients cannot do that, therefore, the JWT is used for authentication, removing the need of the seed. It means that you need to pass the JWT token as the MQTT password and use any username (except empty, since MQTT protocol requires a username to be set if a password is set). The JWT has to have the `Bearer` boolean set to true, which can be done with nsc:
```
kozlovic, line 107: > therefore, the JWT is used for authentication, removing the need of the seed.

It could be me misreading this, but I would disagree with this statement as it is. With a JWT the server does require nonce signing by the client, except if the JWT has the "bearer token"  field enabled. So it is important to refer to the bearer token early on in this explanation. Would the following be better?

In operator mode, all users need to provide a JWT in order to connect, and NATS clients are required to sign a `nonce` sent by the server using their private key \(see [JWTs and Privacy](../securing_nats/jwt/#jwts-and-privacy)\), unless the JWT has the `Bearer` token enabled, which can be done with `nsc` like this:
```
nsc edit user --name U --account A --bearer
```
Since MQTT clients cannot perform the `nonce` signing, they must have a JWT with bearer token. The JWT is provided through the MQTT password. Note that a non empty username still need to be provided since MQTT protocol requires a username to be set if a password is set.

mullerch: I guess the interpretation is a matter of perspective. As a user, I don't know how it's done I just know that secure systems have an identification and an authentication phase. Whether authentication is disabled or done with the JWT is just a point of view.

Maybe it's better (or more honest) to say that authentication is disabled, as the JWT is described everywhere as non-secret.

Your description is fine to me, most importantly the reader should now that the JWT sole is enough to connect through MQTT when the Bearer boolean is set.

(not sure if I should edit something to this PR or if you can change it)

kozlovic (approved):
LGTM!

kozlovic:
@gcolliso I have reviewed and approved. Thanks

gcolliso:
thanks @mullerch

PR #269, opened 2021-06-24
Update queues.md

hi008:
I assume the queue group  still has 2 members or I missing something.


gcolliso:
@hi008 you should have 3 queue groups through steps #3, #4, & #5.

hi008:
Opps, missed the first Go client.

PR #273, opened 2021-06-30
document pinned_certs

matthiashanel:


philpennock (approved):
Should the rows be sorted alphabetically?

matthiashanel:
> Should the rows be sorted alphabetically?

not needed, it will be turned into a map for a point lookup

PR #292, opened 2021-08-25
Update js example to specify the cred file

zwmcoder:
## Description
As from example for js client lib:
https://github.com/nats-io/nats.js/blob/master/examples/nats-pub.js
And the connect ops definition:
https://github.com/nats-io/nats.js/blob/9d53eda057472aceb5b5af9451f2cb7a0d5d033b/index.d.ts#L54

The example here is out of date and not correct. Need update

### DIFF
--- a/developing-with-nats/security/creds.md
+++ b/developing-with-nats/security/creds.md
@@ -50,11 +50,10 @@ nc.close();
 {% tab title="JavaScript" %}
 ```javascript
 // credentials file contains the JWT and the secret signing key
-let credsFile = path.join(confDir, 'credsfile.creds');
-
-let nc = NATS.connect({
-    url: server.nats,
-    userCreds: credsFile
+const data = fs.readFileSync(creds_file_path);
+  let nc = NATS.connect({
+    servers: [url],
+    authenticator: credsAuthenticator(data)
 });
 ```
 {% endtab %}
@@ -80,11 +79,11 @@ await nc.close()
 {% tab title="TypeScript" %}
 ```typescript
 // credentials file contains the JWT and the secret signing key
-let credsFile = path.join(confDir, 'credsfile.creds');
+const data = fs.readFileSync(creds_file_path);
 
 let nc = await connect({
-    url: server.nats,
-    userCreds: credsFile
+    servers: [url],
+    authenticator: credsAuthenticator(data)
 });
 ```
 {% endtab %}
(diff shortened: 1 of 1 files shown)


gcolliso:
@aricart could you review, please? Thank you!

developing-with-nats/security/creds.md
```diff
@@ -50,11 +50,10 @@ nc.close();
 {% tab title="JavaScript" %}
 ```javascript
 // credentials file contains the JWT and the secret signing key
-let credsFile = path.join(confDir, 'credsfile.creds');
-
-let nc = NATS.connect({
-    url: server.nats,
-    userCreds: credsFile
+const data = fs.readFileSync(creds_file_path);
+  let nc = NATS.connect({
```
bruth, line 54:
```suggestion
const data = fs.readFileSync(credsFile);
let nc = NATS.connect({
```

developing-with-nats/security/creds.md
```diff
@@ -80,11 +79,11 @@ await nc.close()
 {% tab title="TypeScript" %}
 ```typescript
 // credentials file contains the JWT and the secret signing key
-let credsFile = path.join(confDir, 'credsfile.creds');
+const data = fs.readFileSync(creds_file_path);
```
bruth, line 82:
```suggestion
const data = fs.readFileSync(credsFile);
```

bruth (changes requested):
Minor code tweaks (naming and indent), but otherwise looks good.

developing-with-nats/security/creds.md
aricart: I would expand on the comment to say that reading of the credentials file will be different depending on the JavaScript runtime, and that the code shown there is for node.js

aricart (approved):
LGTM

developing-with-nats/security/creds.md
```diff
@@ -80,11 +79,11 @@ await nc.close()
 {% tab title="TypeScript" %}
 ```typescript
 // credentials file contains the JWT and the secret signing key
```
bruth, line 81:
```suggestion
// Credentials file contains the JWT and the secret signing key. This examples
// show how the file could be read using the Node.js runtime.
```

bruth:
Merge the commits, but I now see there is a conflict since this is an older PR. Will look shortly.

aricart:
@bruth  I want to also point out that these snippets which we copied right into the doc were actually managed, but when we did the change, we eliminated the CI checking on the sample code.

https://github.com/nats-io/node-nats-examples

PR #293, opened 2021-08-31
Delete prometheus-and-nats-operator.md

wallyqs:


gcolliso:
@wallyqs why are you deleting the file?

wallyqs:
We are cleaning up references to the operator so that Helm charts are preferred instead.

gcolliso:
lgtm

PR #295, opened 2021-09-01
Add nats helm chart README to docs

wallyqs:


wallyqs:
ok added a brief note at the top of the text, they are better because they are easier to extend, maintain and accept contributions.

PR #298, opened 2021-09-09
Add `mqtt_client` in connz output

kozlovic:
Connection information now includes `mqtt_client` for MQTT clients
and this will be the connection's client ID. A filter named `mqtt_client`
can be used to select this connection.

Signed-off-by: Ivan Kozlovic <ivan@synadia.com>


gcolliso:
lgtm - thanks @kozlovic

PR #299, opened 2021-09-13
Update configuration docs to use more precise units

coder543:
In reviewing the configuration docs, I was confused about what values were actually intended. I had go to look at the source code [over here](https://github.com/nats-io/nats-server/blob/b3c19b9dd19171b58da5b5135491848cf72da599/server/const.go#L74) to understand better, so I opened this PR in the hopes that future readers will not feel the need to go source diving, and just to open the conversation around consistent units in the docs.

`Mb` is "megabits", not "megabytes", and just plain `MB` for megabytes would imply base-10... but the values in `const.go` are base-2, so `MiB` is the best unit to use there.

`64K` is ambiguous, and I guessed wrongly that it would be 64000... when that is also base-2 as 65536.

### DIFF
--- a/nats-server/configuration/README.md
+++ b/nats-server/configuration/README.md
@@ -134,10 +134,10 @@ authorization: {
 
 | Property | Description | Default |
 | :--- | :--- | :--- |
-| `max_connections` | Maximum number of active client connections. | `64K` |
-| `max_control_line` | Maximum length of a protocol line \(including combined length of subject and queue group\). Increasing this value may require [client changes](../../developing-with-nats/connecting/misc.md#set-the-maximum-control-line-size) to be used. Applies to all traffic. | `4Kb` |
-| `max_payload` | Maximum number of bytes in a message payload. Reducing this size may force you to implement [chunking](../../developing-with-nats/connecting/misc.md#get-the-maximum-payload-size) in your clients. Applies to client and leafnode payloads. It is not recommended to use values over 8Mb but `max_payload` can be set up to 64Mb. | `1Mb` |
-| `max_pending` | Maximum number of bytes buffered for a connection Applies to client connections. | `64Mb` |
+| `max_connections` | Maximum number of active client connections. | `65536` |
+| `max_control_line` | Maximum length of a protocol line \(including combined length of subject and queue group\). Increasing this value may require [client changes](../../developing-with-nats/connecting/misc.md#set-the-maximum-control-line-size) to be used. Applies to all traffic. | `4KiB` |
+| `max_payload` | Maximum number of bytes in a message payload. Reducing this size may force you to implement [chunking](../../developing-with-nats/connecting/misc.md#get-the-maximum-payload-size) in your clients. Applies to client and leafnode payloads. It is not recommended to use values over 8MiB but `max_payload` can be set up to 64MiB. | `1MiB` |
+| `max_pending` | Maximum number of bytes buffered for a connection Applies to client connections. | `64MiB` |
 | `max_subscriptions` | Maximum numbers of subscriptions per client and leafnode accounts connection. | `0`, unlimited |
 
 ### Authentication and Authorization


gcolliso:
@kozlovic would you be able to review, please?

nats-server/configuration/README.md
```diff
@@ -134,10 +134,10 @@ authorization: {
...
-| `max_control_line` | Maximum length of a protocol line \(including combined length of subject and queue group\). Increasing this value may require [client changes](../../developing-with-nats/connecting/misc.md#set-the-maximum-control-line-size) to be used. Applies to all traffic. | `4Kb` |
-| `max_payload` | Maximum number of bytes in a message payload. Reducing this size may force you to implement [chunking](../../developing-with-nats/connecting/misc.md#get-the-maximum-payload-size) in your clients. Applies to client and leafnode payloads. It is not recommended to use values over 8Mb but `max_payload` can be set up to 64Mb. | `1Mb` |
-| `max_pending` | Maximum number of bytes buffered for a connection Applies to client connections. | `64Mb` |
+| `max_connections` | Maximum number of active client connections. | `65536` |
+| `max_control_line` | Maximum length of a protocol line \(including combined length of subject and queue group\). Increasing this value may require [client changes](../../developing-with-nats/connecting/misc.md#set-the-maximum-control-line-size) to be used. Applies to all traffic. | `4KiB` |
```
kozlovic, line 138: No, it is base 10, so maybe you are right that Kb is ambiguous, but then it is 4KB (4 * 1024 * 1024 = 4194304 bytes), not 4Kib.

nats-server/configuration/README.md
```diff
@@ -134,10 +134,10 @@ authorization: {
...
+| `max_payload` | Maximum number of bytes in a message payload. Reducing this size may force you to implement [chunking](../../developing-with-nats/connecting/misc.md#get-the-maximum-payload-size) in your clients. Applies to client and leafnode payloads. It is not recommended to use values over 8MiB but `max_payload` can be set up to 64MiB. | `1MiB` |
```
kozlovic, line 139: 8MB, 64MB and 1MB

kozlovic (changes requested):
Agreed that maybe we should either use `mb` or `MB` (not `Mb`), but the intent in the doc is kilo bytes, mega bytes. Only probably in few places (like JetStream rate limit) the unit would be in bits. So I agree that we could change to KB/MB, but not Kib/MiB.

nats-server/configuration/README.md
coder543: Base 10 would be 4 * 1000 * 1000. This is base-2, which is why you're multiplying by 1024.

From Wikipedia, here is a unit chart: 
image

`MiB` is the one that uses 1024.

coder543: just based on SI definitions, lowercase `mb` would mean "milli-bit`, which is not a unit anyone should ever use.

kozlovic: Ok, you got me.. But from your same link you see the Memory column that says MB... so that's the one we want.

kozlovic: In other words, we want to use the "Memory" column units, which is KB, MB, GB, TB. This is how most of the time memory usage is reported.

coder543: It's _accepted_ to do MB in memory related contexts, but it is still ambiguous... MiB is explicit that it is powers of 2, and can't be confused for powers of 10.

Regardless, this isn't my docs, so... you can do what you want. I was just trying to help alleviate future confusion for consumers of the docs.

kozlovic: Appreciate the comment and trying to make the doc better, but I just think that MB is more commonly used (and understood) that MiB. Not sure what machine/OS you are running, but I have a mac, if I check "About" and check storage or memory, it uses unit as MB/GB, etc.. and I am sure others do too, not specific to Apple.

You are certainly correct, but to be frank, I am sure there would be more people confused by MiB than MB...

coder543: Yep, definitely pros and cons to both approaches. The ideal solution would be to allow the user to hover over the unit and see some alt-text that says this is base-2 or binary or something, but that's probably not easy with git book.

jnmoyne:
Closing this PR as I have made the changes in the new version of the docs (which is upcoming), the change being in that table from 'Kb' to 'KB' and from 'Mb' to 'MB' (the rest of the docs looks to be pretty consistent about using the appropriate KB/MB notation for memory sizes.

PR #308, opened 2021-09-19
added a hint on running it from docker

aricart:


gcolliso (commented):
lgtm - thanks @aricart

PR #309, opened 2021-09-28
Update docker NATS cluster instructions

ellispritchard:
I'm new to NATS, but this guide didn't work for me, and seems outdated:
* the server needs cluster options to be specified in order to work in cluster mode.
* the name for the cluster needs to be specified to avoid error messages on start-up.
* the port for the monitoring server needs to be specified else it is not started by the server.
* `docker-compose` is now `docker compose`
* `synadia/nats-box` has been updated, and `nats-pub` and `nats-sub` are replaced with the `nats` command.

The above works for me, but please test!

### DIFF
--- a/nats-server/nats_docker/README.md
+++ b/nats-server/nats_docker/README.md
@@ -36,21 +36,24 @@ $ docker network create nats
 ```
 
 ```bash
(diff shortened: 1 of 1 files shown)


gcolliso:
@wallyqs could you review, please? Thank you!

jnmoyne (approved):
LGTM

PR #318, opened 2021-10-30
Fix docs

ColinSullivan1:
Revert back a few revisions.


gcolliso:
LGTM - The files look like they are back to the way they were prior

PR #335, opened 2021-11-28
Fix formatting for nats cli command

boris-ilijic:


gcolliso (approved):
lgmt - thanks @boris-ilijic

PR #340, opened 2021-12-03
Incorrect download link

Shoothzj:
- Fix incorrect download link
- Bump download version to `2.6.5`

### DIFF
--- a/running-a-nats-service/installation.md
+++ b/running-a-nats-service/installation.md
@@ -71,10 +71,10 @@ yay nats-server
 
 You can find the latest release of nats-server [here](https://github.com/nats-io/nats-server/releases/latest).
 
-Download the zip file matching your systems architecture, and unzip. For this example, assuming version 2.6.2 of the server and a Linux AMD64:
+Download the zip file matching your systems architecture, and unzip. For this example, assuming version 2.6.5 of the server and a Linux AMD64:
 
 ```shell
-curl -L https://github.com/nats-io/nats-server/releases/download/v2.0.0/nats-server-v2.6.2-linux-amd64.zip -o nats-server.zip
+curl -L https://github.com/nats-io/nats-server/releases/download/v2.6.5/nats-server-v2.6.5-linux-amd64.zip -o nats-server.zip
 ```
 
 ```shell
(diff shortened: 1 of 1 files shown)


Shoothzj:
@gcolliso PTAL

jnmoyne (approved):
OK

gcolliso:
@jnmoyne this one is difficult as we are pushing versions often and updating this code each time seems inefficient - thoughts?

jnmoyne:
Yeah I agree, in my current (uncommitted) set of changes I have it as:

```
## Downloading a Release Build

You can find the latest release of nats-server on [the nats-io/nats-server GitHub releases page](https://github.com/nats-io/nats-server/releases/).

From the releases page, copy the link to the release archive file of your choice and download it using `curl -L`. 
```
and then below that is the example:

```
## Downloading a Release Build

You can find the latest release of nats-server on [the nats-io/nats-server GitHub releases page](https://github.com/nats-io/nats-server/releases/).

From the releases page, copy the link to the release archive file of your choice and download it using `curl -L`. For this example, assuming version 2.6.2 of the server and a Linux AMD64:
... and it goes on to give the curl line and the output, etc... for in this case version 2.6.2.

My idea was that the first part of the text is 'release version agnostic' with the link being `https://github.com/nats-io/nats-server/releases/` (no version number in that link).

and then only the second part (the 'for example' part) shows some specific version, so of course it gets 'obsolete' as soon as we release a new version (which is pretty often) but it doesn't matter as much IMHO because it's just an example.

**Comment by jnmoyne** [comment]
Actually I have a better idea: using my version for the first part and making the example generic by using X Y Z so people can still copy the whole command and then just edit the X Y and Z to whatever is the current latest release.

I would recommend closing this PR without merging it and I will create a PR from my fork where I will implement this generalization rather than merge this one and then I have to resolve the conflict in my fork. Thank you @Shoothzj for your contribution as it has prompted us to re-think that part of the doc in a better way.

For reference this is what the version of the section in question is now in my branch:

--- 

## Downloading a Release Build

You can find the latest release of nats-server on [the nats-io/nats-server GitHub releases page](https://github.com/nats-io/nats-server/releases/).

From the releases page, copy the link to the release archive file of your choice and download it using `curl -L`.

For example, assuming version X.Y.Z of the server and a Linux AMD64:

```shell

curl -L https://github.com/nats-io/nats-server/releases/download/vX.Y.Z/nats-server-vX.Y.Z-linux-amd64.zip -o nats-server.zip
```

```shell
unzip nats-server.zip -d nats-server
```
which should output something like
```shell
Archive:  nats-server.zip
   creating: nats-server-vX.Y.Z-linux-amd64/
...
```
and finally:
```shell
sudo cp nats-server/nats-server-vX.Y.Z-linux-amd64/nats-server /usr/bin
```

Shoothzj:
> 

You are welcome. I am happy to help with project. I will close this PR. :)

PR #358, opened 2022-01-06
Removed note referring to JetStream pre-cluster support

tbeets:
There was an obsolete note under Storage Overhead section that references _JetStream pre-cluster support_.  Removed.


gcolliso (approved):
lgtm

PR #374, opened 2022-02-14
Update listen address to 0.0.0.0

dirkjot:
On Ubuntu 20.04, the suggested configuration did not work because publishing to `localhost` does not allow external services to reach NATS.  Replaced with `0.0.0.0`.


gcolliso:
@jnmoyne could you please do a quick review? Thank you!

jnmoyne (approved):
Fair enough, approved

PR #376, opened 2022-02-17
fix mixup of discard behavior for MaxBytes vs. MaxMsgs

alanfairless:


gcolliso (approved):
lgtm - thanks @alanfairless

PR #384, opened 2022-03-09
Fix MAX_DELIVERIES subject for JS Event

boris-ilijic:
Subject `JS.EVENT.ADVISORY.MAX_DELIVERIES.<stream>.<consumer>` is not in use, so it is replaces with correct one.

That can be also seen in source here:
https://github.com/nats-io/nats-server/blob/main/server/jetstream_api.go#L204


gcolliso (approved):
lgtm - thanks @boris-ilijic

PR #386, opened 2022-03-22
$JS.API.CONSUMER.DURABLE.CREATE corrected to two token arguments.

tbeets:
As reported in Slack community, line 83 should reflect `$JS.API.CONSUMER.DURABLE.CREATE.*.*` indicating two token parameters.


gcolliso (approved):
lgtm - thanks @tbeets

PR #387, opened 2022-03-22
Fix typo in an account import example

tbeets:
In the text we are giving an example of mapping "foo" in the exporting account to "bar" in the importing account, but the example showed "blo" in the exporting account.


gcolliso (approved):
lgtm - thanks @tbeets

PR #390, opened 2022-03-24
Remove redundant link blob in Connecting page

majoyal:
This is to fix this small issue in the Connecting docs page ([docs.nats.io](https://docs.nats.io/using-nats/developer/connecting)):
image


gcolliso (approved):
lgtm - thanks @majoyal

PR #392, opened 2022-03-30
reference ArtifactHub for Helm configuration

caleblloyd:
Helm chart documentation is automatically uploaded to ArtifactHub every time a release is cut.  Link to ArtifactHub instead of presenting potentially stale information on the NATS docs website.

### DIFF
--- a/running-a-nats-service/nats-on-kubernetes/helm-charts.md
+++ b/running-a-nats-service/nats-on-kubernetes/helm-charts.md
@@ -7,13 +7,15 @@ helm repo add nats https://nats-io.github.io/k8s/helm/charts/
 helm install my-nats nats/nats
 ```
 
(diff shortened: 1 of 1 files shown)


caleblloyd:
It may make more sense to eliminate this page altogether and put the link to ArtifactHub at the bottom of this page:

https://docs.nats.io/running-a-nats-service/nats-kubernetes

gcolliso:
IMO I think we should keep the page that allows it to be on the left menu and possibly easier to find if you are just looking through.

wallyqs (approved):
sounds good to me to point to artifacthub instead

caleblloyd:
Linking off-site is not great for the GitBook format...  I will reflect the most up-to-date Helm configuration here but also preserve the link to ArtifactHub

caleblloyd:
Updated; please re-review

gcolliso:
lgtm

PR #393, opened 2022-03-31
Quote more consistently in protocol docs

Rauno56:
Noticed invalid rendering of "space" and "tab". Fixed those and then noticed the doc usually uses backticks for quoting literal values - made that consistent as well.

Should INBOX be quoted as well? The occurrence is visible in the diff.

## Commits

- fix(protocol): print a space and a tab in "Field Delimiters"
- fix: quote consistently
- fix: write with instead of w/


gcolliso (approved):
@Rauno56 thanks! lgtm

Rauno56:
I noticed that the SSG we are using is removing whitespace from between backticks and omitting empty code blocks after so it still renders wrong. The same issue in the next paragraph:

> [...] followed by `LF` (`CR+LF`, `\r`, `0x0D0A`) to terminate [...]

Issue #400, opened 2022-04-14
Need to dquote in request/reply walkthrough


xieyuschen:
Hi guys, I'm new to the nats-io so I tried nats with the documents. When I read [this page](https://docs.nats.io/nats-concepts/core-nats/reqreply/reqreply_walkthrough) and try the command on the page like this:
```shell
nats reply help.please "OK, I CAN HELP!!!"
```
Then I need to solve the `dquote` problem and I think it will disturb newer who wants learn nats.
<img width="760" alt="image" src="https://user-images.githubusercontent.com/52945328/163712018-92651574-3903-4815-8bda-26478d82b0a7.png">

After I revise `"` to `'`, it works well. 
<img width="915" alt="image" src="https://user-images.githubusercontent.com/52945328/163428647-9098238b-22c4-438f-ae76-be1b5bfb5714.png">

I think it's worth creating a PR to solve it and I will do it. Note that I'm working on the macOs

Thanks

PR #401, opened 2022-04-14
Avoid solving dquote problem in request/reply walkthrough, #400

xieyuschen:
The detail is in issue #400 
Thanks :)


xieyuschen:
Hi @gcolliso , could you review my pr which fixes the command in the document? 
As `!` is a special character in Linux and macOs, so triple `!` will cause a `dquote` event for user to solve.

matthiashanel (approved):
LGTM

matthiashanel:
@xieyuschen thanks for your contribution

PR #407, opened 2022-04-20
Revise to seperate output and command

xieyuschen:
For now, our command and output are combined together like this, it's easy to miss this one command. So I revised it and separate them.
image

Thanks for reviewing:)
@matthiashanel


matthiashanel (approved):
LGTM, Thanks for helping with our docs

PR #409, opened 2022-04-22
Fix the operator name

xieyuschen:
The old one:
> Let’s create an operator called `Operator`:
```bash
nsc add operator MyOperator
```
It should be `MyOperator`, instead of `Operator`.
Thanks, @matthiashanel


matthiashanel (approved):
LGTM, Thanks for improving the docs!

PR #410, opened 2022-04-24
FAQ : Client / Server Version Compatibility

codesplode:
Add a question to handle client server version compatibility. I think NATS' compatibility model is awesome and this handles a lot of the major concerns.


codesplode:
Additional Note regarding compatibility....

If possible and project maintainers agree, I do think many developers including myself would like a simple rule to help guarantee compatibility in apps in addition to this, even if it suggests a more limited compatibility overall. This would also allow NATS to keep its broad compatibility, but be able to handle unforeseen future features that might be more complex or costly to feature flag in the protocol or allow you to cleanup and deprecate / remove some of those signals over time. Here's an example / template of what I might suggest adding to the beginning or end of this content to keep the benefits of this model, while also providing a simple rule:

>For those looking for a simple guarantee regarding NATS version compatibility, we recommend keeping NATS client and server versions within the same major version when possible for the greatest feature parity / stability guarantee (client 2.x.y / server 2.a.b). The NATS team can only guarantee this with NATS maintained clients of course.

PR #414, opened 2022-04-26
MQTT: Add note about server_name

kozlovic:
A server name need to be specified. Give an example on how to configure it.


gcolliso (approved):
lgtm - thanks @kozlovic

PR #418, opened 2022-04-29
JetStream Disaster Recovery revised and updated for clustered mode and latest NATS CLI backup

tbeets:
Simplified and updated JetStream Disaster Recovery documentation.  Now reflects clustered streams with replicas and cleans up some NATS CLI backup examples that were outdated.


gcolliso (approved):
lgtm - thanks @tbeets

jnmoyne:
Sorry a bit late on reviewing this, looks good but one comment is that it would have been nice to have a bit of a walkthrough/example of how to replace a bad node with another (so showing in something like nats stream report or nats stream info the 'replicas' section and explaining what the * or ! in front of the node name means, and then showing using peer-remove, and then the replacement node being started, etc...)

PR #432, opened 2022-05-17
Adjust 'subject mapping' to 'subject mapping and partitioning' to make the partitioning feature more visible as people have trouble finding it in the documentation

jnmoyne:


gcolliso (approved):
lgtm

PR #440, opened 2022-06-07
update docs to match natsbox

solokia:
Just started on nats so was going through the guide and realised nats-sub and nats-pub was giving /bin/sh: not found error.


bruth (approved):
I believe both are still supported, but the NATS CLI is indeed the preferred way going forward.

PR #442, opened 2022-06-16
JNM edits

jnmoyne:
Improve: visibility of JWT in depth guide, how/where to define subject mappings
Remove 'ephemerals can only be push'
New 'enviornmental considerations' page


running-a-nats-service/configuration/configuring_subject_mapping.md
```diff
@@ -4,6 +4,16 @@ _Supported since NATS Server version 2.2_
 
 Subject mapping is a very powerful feature of the NATS server, useful for canary deployments, A/B testing, chaos testing, and migrating to a new subject namespace.
 
+## Configuring subject mapping
+
+Subject mappings are defined and applied at the account level. So if you are using static account security in the server configuration file, or if you are using JWT security in the account JWTs.
```
derekcollison, line 9: "So if" sounds like you are going to followup but does not.

Maybe "if you are using static account security you will need to edit the server configuration file, however if you are using JWT Security (Operator Mode), then you need to use nsc or customer tools to edit and push changes to you account"

jnmoyne:
Thanks for the review and the comments, which should all be addressed by the new commit

derekcollison (approved):
LGTM

PR #447, opened 2022-06-28
Update link to JWT lib, update example and add link to Byron's version of the example (that came out of the original slack thread that caused me to add this info to the docs) and small improvements

jnmoyne:


gcolliso (approved):
lgtm - thanks

PR #453, opened 2022-07-21
fixed link to subject limits in faq

mschneider82:
the link is broken
https://docs.nats.io/reference/faq#does-nats-impose-any-limits-on-the--of-subjects 

should be:
https://docs.nats.io/reference/faq#does-nats-impose-any-limits-on-the-of-subjects


gcolliso (approved):
lgtm - thanks @mschneider82

PR #454, opened 2022-07-21
'exactly once semantics' (i.e. for consumption more than just delivery) is more accurate than 'exactly once delivery'

jnmoyne:
(see slack thread: https://natsio.slack.com/archives/CM3T6T7JQ/p1656590538557769)


gcolliso (approved):
lgtm

PR #460, opened 2022-08-02
Updated the nats unsubscribe to match the nats-py library

arjungandhi:
as per this example in the [nats-py](https://github.com/nats-io/nats.py) repo unsubscribe works slightly differently now. This PR just updates the documentation

```python
    # Simple publisher and async subscriber via coroutine.
    sub = await nc.subscribe("foo", cb=message_handler)

    # Stop receiving after 2 messages.
    await sub.unsubscribe(limit=2)
```


gcolliso:
@wallyqs could you review and approve, please?

PR #463, opened 2022-08-06
Remove duplicate paragraph

seandlg:
As stated, it is suggested to remove the following duplicate paragraph:

> A client will receive a message for each matching subscription, so if a connection has multiple subscriptions using identical or overlapping subjects \(say `foo` and `>`\) the same message will be sent to the client multiple times.


gcolliso (approved):
lgtm - thanks @seandlg for the catch and update.

PR #466, opened 2022-08-11
Fixed routes configuration

matejbizjak:
= instead of :


bruth:
Hi @matejbizjak, thanks for taking the time to make the change, but both `=` and `:` are supported as delimiters in configuration files :wink:

PR #471, opened 2022-08-18
Fix broken link in natsbench page

mprimi:
Current link is broken, see: https://docs.nats.io/using-nats/nats-tools/nats_cli/natsbench
Points to: https://github.com/nats-io/nats.docs/blob/master/using-nats/nats-tools/nats-tools/natscli.md (no such page)

There are multiple ways to link to the proper page, I choose the shortest but let me know if you prefer a different path.


mprimi:
Check failed, but I'm getting 404 from Vercel if I try to inspect, so not sure what is wrong

gcolliso:
@mprimi Thanks for the update. The Vercel error isn't relevant, yet. We are in the process of moving docs from GitBook to next.js site running on Vercel.

PR #476, opened 2022-08-30
fix typo nats-sub to nats sub

Doslin:
1. change nats-sub to nats sub
2. change nats-pub to nats pub


gcolliso:
@wallyqs could you review and approve, please? Thank you!

PR #489, opened 2022-09-19
Add remaining stream config fields

bruth:
In addition, the version column was introduced to indicate the minimum server version required.

The stream name section was removed and incorporated in the table to remove redundancy.


gcolliso (approved):
lgtm

PR #492, opened 2022-09-20
Fix Three Server Example

rodrigc:
Make cluster port match in CLI example and seed.conf file


jnmoyne (approved):
LGTM
Thanks for fixing this!

PR #493, opened 2022-09-21
Improve and fix JS clustering walkthrough

jnmoyne:


gcolliso (approved):
lgtm

conbrad:
Is the functional change these docs refer to backwards compatible? Can one configure a cluster without `accounts` and still connect to it?

PR #504, opened 2022-10-06
Update NATS client protocol with HMSG and HPUB and some CR+LF corrections

tbeets:
This PR adds documentation for the HMSG and HPUB protocol options that were introduced with NATS headers.  Also, some add back and housekeeping of CR+LF (\r\n) protocol termination sequence where had  been inadvertently truncated in previous commits.

### DIFF
--- a/reference/nats-protocol/nats-protocol/README.md
+++ b/reference/nats-protocol/nats-protocol/README.md
@@ -14,7 +14,7 @@ The NATS server implements a [zero allocation byte parser](https://youtu.be/ylRK
 
 **Field Delimiters**: The fields of NATS protocol messages are delimited by whitespace characters ` `(space) or `	`(tab). Multiple whitespace characters will be treated as a single field delimiter.
 
-**Newlines**: NATS uses `CR` followed by `LF` (`CR+LF`, `\r\n`, `0x0D0A`) to terminate protocol messages. This newline sequence is also used to mark the end of the message payload in `PUB`, `MSG`, `HPUB`, and `HMSG` protocol messages.
+**Newlines**: NATS uses ␍ followed by ␊ (␍␊, `0x0D0A`) to terminate protocol messages. This newline sequence is also used to mark the end of the message payload in `PUB`, `MSG`, `HPUB`, and `HMSG` protocol messages.
 
 **Subject names**: Subject names, including reply subject (INBOX) names, are case-sensitive and must be non-empty alphanumeric strings with no embedded whitespace. All ascii alphanumeric characters except spaces/tabs and separators which are `.` and `>` are allowed. Subject names can be optionally token-delimited using the dot character (`.`), e.g.:
 
--- a/reference/nats-protocol/nats-protocol/README.md
+++ b/reference/nats-protocol/nats-protocol/README.md
@@ -14,7 +14,7 @@ The NATS server implements a [zero allocation byte parser](https://youtu.be/ylRK
 
 **Field Delimiters**: The fields of NATS protocol messages are delimited by whitespace characters ` `(space) or `	`(tab). Multiple whitespace characters will be treated as a single field delimiter.
 
-**Newlines**: NATS uses ␍ followed by ␊ (␍␊, `0x0D0A`) to terminate protocol messages. This newline sequence is also used to mark the end of the message payload in `PUB`, `MSG`, `HPUB`, and `HMSG` protocol messages.
+**Newlines**: NATS uses `␍` followed by `␊` (`␍␊`, `0x0D0A`) to terminate protocol messages. This newline sequence is also used to mark the end of the message payload in `PUB`, `MSG`, `HPUB`, and `HMSG` protocol messages.
 
 **Subject names**: Subject names, including reply subject (INBOX) names, are case-sensitive and must be non-empty alphanumeric strings with no embedded whitespace. All ascii alphanumeric characters except spaces/tabs and separators which are `.` and `>` are allowed. Subject names can be optionally token-delimited using the dot character (`.`), e.g.:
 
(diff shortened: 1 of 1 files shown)


tbeets:
Hmm. Confirmed that GitBooks rendered content is truncating every single instance of `\r\n` into `\r`.

reference/nats-protocol/nats-protocol/README.md
```diff
@@ -162,15 +164,54 @@ where:
...
 To publish an empty message to subject NOTIFY:
 
-`PUB NOTIFY 0\r\n\r`
+`PUB NOTIFY 0␍␊␍␊`
+
+## HPUB
+
+### Description
+
+The `HPUB` message is the same as `PUB` but extends total message payload to include NATS headers. Note that the payload itself is optional. To omit the payload, set the total message size to the same as the headers size. Note that the trailing CR+LF is still required.
```
bruth, line 181:
```suggestion
The `HPUB` message is the same as `PUB` but extends the message payload to include NATS headers. Note that the payload itself is optional. To omit the payload, set the total message size equal the size of the headers. Note that the trailing CR+LF is still required.
```

reference/nats-protocol/nats-protocol/README.md
```diff
@@ -162,15 +164,54 @@ where:
...
+
+NATS headers are similar in semantic to HTTP headers as `name: value` pairs. Multi-value headers are allowed. Headers can be mixed case and NATS will preserve case between message publisher and message receiver(s).  See also [ADR-4 NATS Message Headers](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-4.md).
```
bruth, line 183:
```suggestion
NATS headers are similar, in structure and semantics, to HTTP headers as `name: value` pairs including supporting multi-value headers. Headers can be mixed case and NATS will preserve case between message publisher and message receiver(s).  See also [ADR-4 NATS Message Headers](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-4.md).
```

reference/nats-protocol/nats-protocol/README.md
```diff
@@ -245,11 +286,41 @@ where:
...
+
+To deliver the same message along with a reply inbox:
+
+`MSG FOO.BAR 9 INBOX.34 11␍␊Hello World␍␊`
+
+## HMSG
+
+### Description
+
+The `HMSG` message is the same as `MSG` but extends message payload with headers. See also [ADR-4 NATS Message Headers](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-4.md).
```
bruth, line 299:
```suggestion
The `HMSG` message is the same as `MSG`, but extends the message payload with headers. See also [ADR-4 NATS Message Headers](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-4.md).
```

bruth (commented):
A couple wording/grammar suggestions, but looks great otherwise!

gcolliso:
I don't really understand the CR/LF stuff but in the doc under Protocol conventions/New lines, they are really small and hard to read. If they are in a code block they render a bit bigger like in the PUB syntax section

tbeets:
I've made the edits suggested by @gcolliso and @bruth .

reference/nats-protocol/nats-protocol/README.md
```diff
@@ -162,15 +164,54 @@ where:
...
 To publish an empty message to subject NOTIFY:
 
-`PUB NOTIFY 0\r\n\r`
+`PUB NOTIFY 0␍␊␍␊`
+
+## HPUB
+
+### Description
+
+The `HPUB` message is the same as `PUB` but extends the message payload to include NATS headers. Note that the payload itself is optional. To omit the payload, set the total message size equal the size of the headers. Note that the trailing CR+LF is still required.
```
bruth, line 181:
```suggestion
The `HPUB` message is the same as `PUB` but extends the message payload to include NATS headers. Note that the payload itself is optional. To omit the payload, set the total message size equal to the size of the headers. Note that the trailing CR+LF is still required.
```

bruth: I think I added this correction before. In case it was overlooked, just calling it out again.. or if the current wording is grammatically correct then that is fine too. It just reads a bit weird for me.

reference/nats-protocol/nats-protocol/README.md
```diff
@@ -162,15 +164,54 @@ where:
...
+NATS headers are similar, in structure and semantics, to HTTP headers as `name: value` pairs including supporting multi-value headers. Headers can be mixed case and NATS will preserve case between message publisher and message receiver(s).  See also [ADR-4 NATS Message Headers](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-4.md).
+
+### Syntax
+
+`HPUB <subject> [reply-to] <#header bytes> <#total bytes>␍␊[headers]␍␊␍␊[payload]␍␊`
+
+where:
+
+* `subject`: The destination subject to publish to
+* `reply-to`: The optional reply inbox subject that subscribers can use to send a response back to the publisher/requestor
```
bruth, line 192: Since an *inbox* subject is not technically required (in the `_INBOX.>` sense of the word).. I would suggest the word be removed since it may be misinterpreted as a requirement.

If you agree, there are a few other places where "inbox" it used that should be updated.

tbeets: I agree.  As long as we've opened up the patient, we can do some additional nips and tucks.

bruth (approved):

gcolliso (approved):
lgtm

PR #506, opened 2022-10-08
Correct docs to say mirror can filter subject

Lemorz56:
The documentation for Replication currently says that a mirror cannot filter by subject.
This was changed for release 2.9.0 and https://github.com/nats-io/nats-server/pull/3227

I've changed the docs to say that a mirror can filter by subject.


gcolliso (approved):
lgtm - thank you for the update @Lemorz56

PR #515, opened 2022-10-18
Break out 'clustering' into its own table, to make it easier to find

rodrigc:
See: https://github.com/nats-io/nats.docs/blob/26cb21f4189365955b824dd5ca36a92445a77769/running-a-nats-service/configuration/README.md#clustering


gcolliso (approved):
lgtm - thanks @rodrigc

PR #530, opened 2022-11-10
Update basics.md

erikjw:
"Incantation" isn't really the correct word in this context.  I propose to change it to "command" here.

I also propose changing "it's" to "its".  It's (with apostrophe) in English is a contraction meaning "it is", whereas the possessive (which I believe is what was intended here) is merely "its" without the apostrophe.


gcolliso (approved):
lgtm - thanks @erikjw

PR #532, opened 2022-11-10
fix broken nats.rs link to TLS

gcolliso:


gcolliso:
addressing Issue #532

PR #549, opened 2022-12-15
nsc is now v2 from Go perspective

philpennock:
As of nsc `v2.7.6`, it presents as v2 to the Go module namespace perspective,
unlike `2.7.5` (no `v` prefix).  So this install command should be updated.


gcolliso (approved):
lgtm

Issue #551, opened 2022-12-28
Add more explainations and settings related to NATS gateway behind a Load Balancer


rguilmont:
### Context

We encountered several issues with nats supercluster when the nodes are behind a loadbalancer. Here is the chat discussion on Slack related to that : [here](https://natsio.slack.com/archives/C069GSYFP/p1670259357636119)

rguilmont:
Closing since the PR has been merged, thanks a lot NATS team !

PR #552, opened 2022-12-28
Add more explainations and settings related to NATS gateway behind a Load Balancer

rguilmont:
Related to #551 

I wasn' t able to serve locally using gitbook-cli, so the link i' ve put might be broken and need to be double checked.

If i can add anything to this, don' t hesitate !


gcolliso:
@wallyqs could you review/approve please? TY

wallyqs (approved):
LGTM, thanks

PR #554, opened 2023-01-06
Add `profiling.md`

neilalexander:
This adds a page on how to capture CPU and memory profiles, in the odd case that it's useful.


gcolliso (approved):
lgtm - thanks @neilalexander

PR #555, opened 2023-01-06
Fix reference to CLI flag `--auth`

homeworkprod:
"token" is the name of the flag's argument, but the name of the flag itself is `--auth`.


bruth (approved):

bruth:
Thanks @homeworkprod!

PR #558, opened 2023-01-17
fix(protocol): add missing headers in HMSG message syntax summary

joanrieu:
The protocol description mentions the _headers_ which are the most important thing about the `HMSG` message, but the syntax summary line was missing them.

I have added them in this MR, like in the `HPUB` message docs.


gcolliso (approved):
lgtm - thanks @joanrieu

PR #560, opened 2023-01-27
Fix mirror and sources links

bruth:
The relative path was being interpreted incorrectly as a relative path to GitHub.


gcolliso:
I fixed this using the Gitbook UI instead of this PR. We need to keep the relative links in Gitbook, but this is my guess why these links aren't working sometimes - GitBook can be sensitive to having the folders named differently in the source code vs what is displayed on the summary.md. So in the code, the folder is nats_admin but in the summary.md it is `Managing and Monitoring your NATS Server Infrastructure`.  JNM did a lot of non-Gitbook standard stuff when he rearranged the docs a while ago. So technically the links work but when the UI of Gitbook tries to sync with the GitHub code it gets confused.

PR #563, opened 2023-02-06
Expand on using environment variables and give example of prompting for a password

jnmoyne:


gcolliso (approved):
lgtm - thanks @jnmoyne

PR #571, opened 2023-02-23
Add reference table for each message showing further details

### LINKED ISSUE #570: Client Protocol Field Documentation
The client protocol documentation is lacking detail for certain fields. Type information, and optionality would be very useful to have at a glance.

https://docs.nats.io/reference/reference-protocols/nats-protocol#connect for e.g. shows a list of fields, some of which are explicitly marked as optional, some with a dependency of `if auth_required is set` (which also begs the question, what if it's set to false?) and some without any indication. However in the example messages, there are fields missing that aren't explicitly called out as optional 

e.g. 

```
CONNECT {"verbose":false,"pedantic":false,"tls_required":false,"name":"","lang":"go","version":"1.2.2","protocol":1}␍␊
```

is missing the field `jwt` but this field is not marked as optional.

Suggest a table format for such fields as below or similar

|name|description|type|required|
|-------|----|---------|------------|
| jwt |The JWT that identifies a user permissions and account.| string | false|
| user |client name| string | if `auth_required` is true |

I'm happy to submit a PR if a change is deemed necessary.

samisagit:
Show name, description, type and optionality of each field in each message.

Closes #570

### DIFF
--- a/reference/nats-protocol/nats-protocol/README.md
+++ b/reference/nats-protocol/nats-protocol/README.md
@@ -71,19 +71,33 @@ When using the updated client protocol (see [`CONNECT`](./#connect) below), `INF
 
 The valid options are as follows, encoded as JSON:
 
-* `server_id`: The unique identifier of the NATS server
-* `version`: The version of the NATS server
-* `go`: The version of golang the NATS server was built with
-* `host`: The IP address used to start the NATS server, by default this will be `0.0.0.0` and can be configured with `-client_advertise host:port`
-* `port`: The port number the NATS server is configured to listen on
-* `max_payload`: Maximum payload size, in bytes, that the server will accept from the client.
-* `proto`: An integer indicating the protocol version of the server. The server version 1.2.0 sets this to `1` to indicate that it supports the "Echo" feature.
-* `client_id`: An optional unsigned integer (64 bits) representing the internal client identifier in the server. This can be used to filter client connections in monitoring, correlate with error logs, etc...
-* `auth_required`: If this is set, then the client should try to authenticate upon connect.
-* `tls_required`: If this is set, then the client must perform the TLS/1.2 handshake. Note, this used to be `ssl_required` and has been updated along with the protocol from SSL to TLS.
-* `tls_verify`: If this is set, the client must provide a valid certificate during the TLS handshake.
-* `connect_urls` : An optional list of server urls that a client can connect to.
-* `ldm`: If the server supports _Lame Duck Mode_ notifications, and the current server has transitioned to lame duck, `ldm` will be set to `true`.
+| name              | description                                                                                                                                                            | type     | presence |
+|-------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------|----------|
+| `server_id`       | The unique identifier of the NATS server.                                                                                                                              | string   | always   |
+| `server_name`     | The name of the NATS server.                                                                                                                                           | string   | always   |
+| `version`         | The version of NATS.                                                                                                                                                   | string   | always   |
+| `go`              | The version of golang the NATS server was built with.                                                                                                                  | string   | always   |
+| `host`            | The IP address used to start the NATS server, by default this will be `0.0.0.0` and can be configured with `-client_advertise host:port`.                              | string   | always   |
+| `port`            | The port number the NATS server is configured to listen on.                                                                                                            | int      | always   |
+| `headers`         | Whether the server supports headers.                                                                                                                                   | bool     | always   |
+| `max_payload`     | Maximum payload size, in bytes, that the server will accept from the client.                                                                                           | int      | always   |
+| `proto`           | An integer indicating the protocol version of the server. The server version 1.2.0 sets this to `1` to indicate that it supports the "Echo" feature.                   | int      | always   |
+| `client_id`       | The internal client identifier in the server. This can be used to filter client connections in monitoring, correlate with error logs, etc...                           | uint64   | optional |
+| `auth_required`   | If this is true, then the client should try to authenticate upon connect.                                                                                              | bool     | optional |
+| `tls_required`    | If this is true, then the client must perform the TLS/1.2 handshake. Note, this used to be `ssl_required` and has been updated along with the protocol from SSL to TLS.| bool     | optional |
+| `tls_verify`      | If this is true, the client must provide a valid certificate during the TLS handshake.                                                                                 | bool     | optional |
+| `tls_available`   | If this is true, the client can provide a valid certificate during the TLS handshake.                                                                                  | bool     | optional |
+| `connect_urls`    | List of server urls that a client can connect to.                                                                                                                      | [string] | optional |
+| `ws_connect_urls` | List of server urls that a websocket client can connect to.                                                                                                            | [string] | optional |
+| `ldm`             | If the server supports _Lame Duck Mode_ notifications, and the current server has transitioned to lame duck, `ldm` will be set to `true`.                              | bool     | optional |
+| `git_commit`      | The git hash at which the NATS server was built.                                                                                                                       | string   | optional |
+| `jetstream`       | Whether the server supports JetStream.                                                                                                                                 | bool     | optional |
+| `ip`              | The IP of the server.                                                                                                                                                  | string   | optional |
+| `client_id`       | The ID of the client.                                                                                                                                                  | string   | optional |
+| `client_ip`       | The IP of the client.                                                                                                                                                  | string   | optional |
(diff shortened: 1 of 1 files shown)


reference/nats-protocol/nats-protocol/README.md
```diff
@@ -71,19 +71,22 @@ When using the updated client protocol (see [`CONNECT`](./#connect) below), `INF
...
+|`max_payload`|Maximum payload size, in bytes, that the server will accept from the client.|int|true|
+|`proto`|An integer indicating the protocol version of the server. The server version 1.2.0 sets this to `1` to indicate that it supports the "Echo" feature.|int|true|
+|`client_id`|The internal client identifier in the server. This can be used to filter client connections in monitoring, correlate with error logs, etc...|uint64|false|
+|`auth_required`|If this is set, then the client should try to authenticate upon connect.|bool|false|
+|`tls_required`|If this is set, then the client must perform the TLS/1.2 handshake. Note, this used to be `ssl_required` and has been updated along with the protocol from SSL to TLS.|bool|false|
+|`tls_verify`|If this is set, the client must provide a valid certificate during the TLS handshake.|bool|false|
+|`connect_urls`|List of server urls that a client can connect to.|[string]|false|
+|`ldm`|If the server supports _Lame Duck Mode_ notifications, and the current server has transitioned to lame duck, `ldm` will be set to `true`.|bool|false|
+|`headers`|Whether the server supports headers|bool|false|
```
samisagit, line 89: happy to reformat all of these tables into 'prettier' versions, but would be good to get confirmation on the content before tuning that.

reference/nats-protocol/nats-protocol/README.md
```diff
@@ -71,19 +71,22 @@ When using the updated client protocol (see [`CONNECT`](./#connect) below), `INF
...
-* `port`: The port number the NATS server is configured to listen on
-* `max_payload`: Maximum payload size, in bytes, that the server will accept from the client.
-* `proto`: An integer indicating the protocol version of the server. The server version 1.2.0 sets this to `1` to indicate that it supports the "Echo" feature.
-* `client_id`: An optional unsigned integer (64 bits) representing the internal client identifier in the server. This can be used to filter client connections in monitoring, correlate with error logs, etc...
-* `auth_required`: If this is set, then the client should try to authenticate upon connect.
-* `tls_required`: If this is set, then the client must perform the TLS/1.2 handshake. Note, this used to be `ssl_required` and has been updated along with the protocol from SSL to TLS.
-* `tls_verify`: If this is set, the client must provide a valid certificate during the TLS handshake.
-* `connect_urls` : An optional list of server urls that a client can connect to.
-* `ldm`: If the server supports _Lame Duck Mode_ notifications, and the current server has transitioned to lame duck, `ldm` will be set to `true`.
```
samisagit, line 86: there seems to be a mix of punctuation here, are full stops at the end of each description preferred?

bruth:
This is great, thank you @samisagit. I would err on the side of full stop (period) for consistency.

bruth:
For the messages that are returned by the server, I don't think the `required` column is necessary. Those fields are all included, but may have varying values depending on configuration/state.

samisagit:
> For the messages that are returned by the server, I don't think the `required` column is necessary. Those fields are all included, but may have varying values depending on configuration/state.

Yeah I agree that language isn't really correct - I wonder if a header of `presence` with values of `always` and `optional` would be more correct. I'm thinking of fields like `reply_to` and `payload` that aren't always present.

bruth:
> I wonder if a header of `presence` with values of `always` and `optional`

I agree with using the term `presence`. For the `INFO` message in particular, we can refer to [this](https://github.com/nats-io/nats-server/blob/c1af58507b388b40d90afbd10eb09d83602ba645/server/server.go#L60-L85) which indirectly declares which fields may not be present via the `omitempty` declaration. For client connections specifically,

- `git_commit`
- `auth_required`
- `tls_required`
- `tls_available`
- `tls_verify`
- `jetstream`
- `ip`
- `client_id`
- `client_ip`
- `nonce`
- `cluster`
- `domain`
- `connect_urls`
- `ws_connect_urls`
- `ldm`

> Should these be replaced with "If this is true"

Yes agree with that. "If set" is ambiguous.

reference/nats-protocol/nats-protocol/README.md
```diff
@@ -119,20 +134,26 @@ The `CONNECT` message is the client version of the [`INFO`](./#info) message. On
...
+| `version`       | The version of the client.                                                                                                                                                                                                                                                        | string | true                         |
+| `protocol`      | Sending `0` (or absent) indicates client supports original protocol. Sending `1` indicates that the client supports dynamic reconfiguration of cluster topology changes by asynchronously receiving [`INFO`](./#info) messages with known servers it can reconnect to.            | int    | false                        |
+| `echo`          | If set to `false`, the server (version 1.2.0+) will not send originating messages from this connection to its own subscriptions. Clients should set this to `false` only for server supporting this feature, which is when `proto` in the `INFO` protocol is set to at least `1`. | bool   | false                        |
```
samisagit, line 149: to note - I have flipped the description on this from `true` to `false` since I think it was documented incorrectly. Happy to discuss if necessary.

reference/nats-protocol/nats-protocol/README.md
```diff
@@ -119,20 +134,26 @@ The `CONNECT` message is the client version of the [`INFO`](./#info) message. On
...
+| `nkey`          | ???.                                                                                                                                                                                                                                                                              | string | false                        |
+| `account`       | ???.                                                                                                                                                                                                                                                                              | bool   | false                        |
+| `account_new`   | ???.                                                                                                                                                                                                                                                                              | bool   | false                        |
```
samisagit, line 156: unsure what these fields are used for without diffing into the server code. If someone does know of the top of their head that would be appreciated.

reference/nats-protocol/nats-protocol/README.md
```diff
@@ -71,19 +71,34 @@ When using the updated client protocol (see [`CONNECT`](./#connect) below), `INF
...
+| `client_ip`       | The IP of the client.                                                                                                                                                  | string   | optional |
+| `nonce`           | The nonce for use in CONNECT.                                                                                                                                          | string   | optional |
+| `cluster`         | The name of the cluster.                                                                                                                                               | string   | optional |
+| `cluster_dynamic` | Whether the cluster is dynamic.                                                                                                                                        | bool     | optional |
+| `domain`          | The domain name of the server.                                                                                                                                         | bool     | optional |
```
samisagit, line 101: I have inferred several of the above field descriptions. Let me know if any need altering.

bruth:
```suggestion
| `domain`          | The configured NATS domain of the server.                                                                                                                                         | string     | optional |
```

reference/nats-protocol/nats-protocol/README.md
```diff
@@ -71,19 +71,34 @@ When using the updated client protocol (see [`CONNECT`](./#connect) below), `INF
...
+| `client_id`       | The ID of the client.                                                                                                                                                  | string   | optional |
+| `client_ip`       | The IP of the client.                                                                                                                                                  | string   | optional |
+| `nonce`           | The nonce for use in CONNECT.                                                                                                                                          | string   | optional |
+| `cluster`         | The name of the cluster.                                                                                                                                               | string   | optional |
+| `cluster_dynamic` | Whether the cluster is dynamic.                                                                                                                                        | bool     | optional |
```
bruth, line 100: `cluster_dynamic` is specific to gateway connections if I recall. This can be removed.

reference/nats-protocol/nats-protocol/README.md
bruth: You are correct, default opts on the server show it is set to `true`.

reference/nats-protocol/nats-protocol/README.md
bruth, line 156: We can drop `account` and `account_new`, it is unclear what these are used for, possibly not for client connections or they are deprecated. For example the [Go client](https://github.com/nats-io/nats.go/blob/main/nats.go#L755) does not have these fields as part of the connect struct.

```suggestion
| `nkey`          | The public NKey to authenticate the client. This will be used to verify the signature (`sig`) against the `nonce` provided in the `INFO` message.                                                                                                                                                                                                                                                                            | string | false |
```

samisagit:
Sorry for the delay, life got a bit busy - just pushed some changes that I think the resolve all the feedback. Let me know if there are any further changes required.

bruth (approved):
@samisagit No apology necessary! This is a great contribution and very appreciated.

PR #574, opened 2023-03-13
Add more links to JetStream configuration and code examples

rodrigc:
This page is the first to come up in a web search for "NATS JetStream".
Put more links to configuration and code examples, to improve SEO.

See rendered form of page:
https://github.com/nats-io/nats.docs/blob/bcb15e6771243aaa5d2343e1270a949430c3f23f/nats-concepts/jetstream/readme.md#configuration


jnmoyne (approved):
I like the addition of those links, personally I would have added them at the bottom but fair enough it's ok where they are for the people that want to quickly go to config/example after landing on the concept page from a search engine, so LGTM.

PR #577, opened 2023-03-20
Update link to the memroy session

nomailme:
Section anchors are in lower case, while the hyperlinks at the top of the page are in uppercase and therefore they are not working. Updated hyperlinks to be in lowercase.


gcolliso (approved):
lgtm - thank you @nomailme

PR #578, opened 2023-03-22
Correct the issuer in the simple user creation example

zemocasu:
The "simple user creation" example currently sets the `IssuerAccount` of the new user claim to the public account signing key: https://docs.nats.io/running-a-nats-service/nats_admin/security/jwt#simple-user-creation

This results in a failure when attempting to use these user credentials to connect to the server, like:
```
Account JWT lookup error: account missing
```

Based on the signing key documentation, it looks like the `IssuerAccount` should instead be set to the account's public key: https://docs.nats.io/using-nats/nats-tools/nsc/signing_keys


jnmoyne (approved):
Indeed when using a signing key you also need to add the account's public key so that the server can find (and validate) the signing key. Thanks for the contribution.
LGTM

PR #585, opened 2023-04-21
Backfill STAN minor releases

bruth:
Also add the deprecation notice to the Changes page.


bruth:
@derekcollison Any other specific changes you want to see incorporated? The current page only shows minor releases (no patches), however the release notes on GitHub don't have a release listed for every tag (for example v0.25.0 is not present).

derekcollison (commented):
Should we add in any more details on what was added from 0.10 - 0.25?

bruth:
Backfilled the other minor releases for posterity.

PR #590, opened 2023-05-17
config options leafnode -> leafnodes

caleblloyd:
Singular `leafnode` is not an option, standardize to `leafnodes`

https://github.com/nats-io/nats-server/blob/5db57fb053a03dd84928ed84fb14da9e2267d1eb/server/opts.go#L902-L907


gcolliso (approved):
lgtm - thanks @caleblloyd

PR #597, opened 2023-06-08
fix: typo in leafnode example

EJTJ3:


bruth:
Thanks @EJTJ3. Interestingly (or confusingly enough) `nats-req` was an existing standalone CLI that predated the more encompassing `nats` CLI we know today! But this is definitely a preferred fix for these kinds of examples, so thanks!

bruth (approved):

EJTJ3:
@bruth I didn't know that. I was playing around with leafnode's when I found this little bug/feature :P.

Thnx for approving so quickly.

PR #601, opened 2023-06-13
MQTT Support for `.` character

kozlovic:
Starting at v2.9.10, NATS will support the character `.` in MQTT topics and translate it to `//` in NATS subject.


gcolliso (approved):
lgtm - thanks @kozlovic

PR #604, opened 2023-08-13
Added missing options for jsz, healthz and connz endpoints

mdawar:
While working on a monitoring app, I found that there was missing options for some endpoints, this pull request adds them and also fixes a link for the `jsz` endpoint.

This pull request changes:
- Adds `rtt` sorting option for the `connz` endpoint
- Fixes `jsz` link with `consumers` option (Was `consumer` without `s`)
- Adds `raft` bool option for the `jsz` endpoint
- Adds `js-enabled-only` option for `healthz` and marks `js-enabled` as [deprecated](https://github.com/nats-io/nats-server/blob/f2c7a9d37f1a7a612814abf9365c52ed6687ec4f/server/monitor.go#L2690)


bruth (approved):
LGTM. Thank you!

PR #609, opened 2023-08-17
Update and prune k8s docs

bruth:


bruth:
Need to update the `SUMMARY.md` to remove old links..

caleblloyd (approved):
LGTM

gcolliso:
Considering so many pages are being deleted, should we either check for links from other sections or put some redirects in for the deleted pages?

bruth:
Yes, good call. I will check.

bruth:
Added the redirects

gcolliso (approved):
lgtm -thanks @bruth

PR #610, opened 2023-08-23
Add stream rollup section

bruth:


bruth:
@jnmoyne A user asked about the `Nats-Rollup` behavior and subsequently for an overview of [all the headers](https://docs.nats.io/~/revisions/CpqWB6SZQMLTXjkbWmVj/nats-concepts/jetstream/headers) that are supported. The one set I left empty for now were the mirror-related ones since that seemed to be for internal use. I can remove if that is true.

nats-concepts/jetstream/headers.md
```diff
@@ -0,0 +1,55 @@
...
+| `Nats-Sequence`      | The original sequence of the message.                                                                                  | `193`                         | 2.8.3   |
+| `Nats-Last-Sequence` | The last sequence of the message having the same subject, otherwise zero if this is the first message for the subject. | `190`                         | 2.8.3   |
+
+## Sources
+
+Headers that are implicitly added to messages sourced from other streams.
+
+| Name                 | Description                                                 | Example     | Version |
+| :------------------- | :---------------------------------------------------------- | :---------- | :------ |
+| `Nats-Stream-Source` | Specifies the stream a message sourced from, if applicable. | `my-stream` | 2.2.0   |
```
jnmoyne, line 36: Specifies the original stream name, the subject and the sequence number plus the subject filter and destination transform of the message being sourced.

jnmoyne (approved):
Just the suggestion in the comment otherwise LGTM!

nats-concepts/jetstream/headers.md
bruth: Updated to use this definition.

PR #612, opened 2023-09-02
Improve nsc basics walkthrough

jnmoyne:
Improve the walkthrough to show using nsc to generate the security part of the server config and how to create a nats context with the created credentials


gcolliso (approved):
lgtm

PR #617, opened 2023-09-27
Update installation.md

autodidaddict:
Updates the instructions for installing from source to be current with the version of Go that the server requires.


jnmoyne (approved):
LGTM

PR #629, opened 2023-11-04
Fix command server check

rogerogers:
Fix command server check


bruth (approved):

bruth:
Thanks for the fix @rogerogers. The fix was applied in a previous update.

PR #632, opened 2023-11-14
Correct Syslog and describe used standard

marshauf:
nats-server doesn't recognize `syslog`as a protocol:
```
2023/11/14 10:23:31 error invalid network type: "syslog"
```
It should probably say tcp instead of syslog.

On what I could find, NATS Server sends RFC5424 protocol based messages with non transparent framing.
This information is important for configuring Syslog receivers.

### DIFF
--- a/running-a-nats-service/configuration/logging.md
+++ b/running-a-nats-service/configuration/logging.md
@@ -40,22 +40,23 @@ If `-T false` then log entries are not timestamped. Default is true.
 
 #### Syslog
 
+NATS Server follows the RFC5424 protocol standard and uses non transparent framing.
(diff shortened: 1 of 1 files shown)


marshauf:
The syslog package which NATS Server uses, actually doesn't follow any standard. See https://github.com/nats-io/nats-server/issues/4793#issuecomment-1810388433

`NATS Server follows the RFC5424 protocol standard and uses non transparent framing.`is therefore wrong.

gcolliso:
@bruth should we just update the syslog -> tcp parts and then address any standards once the issue is updated?

PR #641, opened 2024-01-10
Expand on max acks pending for consumers

jnmoyne:
Expand on max acks pending for consumers prompted by the slack thread https://natsio.slack.com/archives/C069GSYFP/p1704671974384789


gcolliso (approved):
lgtm - thanks @jnmoyne

PR #643, opened 2024-01-17
nk: fix installation instructions

jkraml-staffbase:
- `go get` [is deprecated](https://golang.org/doc/go-get-install-deprecation) for installing binaries
- supplying an actual version (instead of `@latest`) would be nice, but that would have to be maintained. Just using the latest version is also fine for the scope of this document, IMHO


gcolliso (approved):
lgtm - thanks @jkraml-staffbase

PR #656, opened 2024-03-19
docs: add k8s Helm and JetStream permission examples

a-h:


bruth:
Thanks @a-h. I will be reviewing this in the next day or two. It prompted me to dig into the concrete permissions required depending on the consumer and stream requirements.

PR #657, opened 2024-03-20
remove word "can"

mantcz:
There is double "can" where only one is sufficient.


gcolliso (approved):
Thanks @mantcz for the update. Made a few other corrections I saw as well.

PR #664, opened 2024-04-18
Remove colon from the unique_tags value

jarretlavallee:
Before this change, the value of the `unique_tag` was in the format of `az:` instead of `az`. The colon causes the unique tag to not function correctly. This change removes the colon and aligns the value with the tests in `nats-server`.


gcolliso (approved):
lgtm

PR #666, opened 2024-04-21
Update walkthrough_setup.md

vaibhavpatil123:


gcolliso:
@ripienaar could you review the addition here, please?

Echo-Head-Wall:
@vaibhavpatil123 you misspelled nats as ants also your 
echo "export PATH=$PATH:~/usr/local/Cellar/nats/0.1.4/bin" >> ~/.zshrc vi changes source ~/.zshrc
should be in the code block. i think having the return of running the command here is not useful as no other example has it and its not mac specific

vaibhavpatil123:
> @vaibhavpatil123 you misspelled nats as ants also your echo "export PATH=$PATH:~/usr/local/Cellar/nats/0.1.4/bin" >> ~/.zshrc vi changes source ~/.zshrc should be in the code block or at least a second separate one. i think having the return of running the command here is not useful as no other example has it and its not mac specific Edit: ( i dont have a mac but i asked a friend who has one to try it and they said that it just worked no need to export or append to zshrc so im kinda confused what this if for)


thanks for input changes are incorporated please review and approve it thanks

Echo-Head-Wall:
i still don't get why you're including the entire output of the command seems unneeded and a lot of bloat. also your steps are version specific and again are you sure for using brew on mac it is required to manually add path since as i mentioned before i had a friend check and it worked fine withought adding it to path manually.

nats-concepts/what-is-nats/walkthrough_setup.md
```diff
@@ -14,6 +14,72 @@ For MacOS:
...
+## Step 1: Add NATS Server to PATH
+
+First, you need to append the NATS server directory to your PATH variable. This can be done by adding the following line to your `.zshrc` file. You can use a text editor like `vi` for this purpose:
+
+```sh
+vi ~/.zshrc
+```
+In the editor, add the following line at the end of the file:
+```
+export PATH=\$PATH:~/usr/local/Cellar/nats/0.1.4/bin
```
ripienaar, line 30: this is not neccesary, the reason you have to do this is because your homebrew setup is incorrect.  Correctly install homebrew and this will just work 

These paths are also very unusual and probably due to something that went wrong with your homebrew

PR #669, opened 2024-04-24
Fixed spelling error of ncs to nsc within streams.md

Echo-Head-Wall:
nsc was accidentally misspelled as ncs this fix's that.


Echo-Head-Wall:
figured the error may be in more then one file so searched for it and found the rest to fix them

PR #670, opened 2024-05-08
Add installing from the shell

jnmoyne:
Add installing from the shell using curl


gcolliso (approved):
lgtm - thanks @jnmoyne

PR #671, opened 2024-05-13
Fixup pubsub walkthrough docs

### LINKED ISSUE #650: example mis-match on pubsub_walkthrough page
Hi,
While going through docs I found that on publisher subscriber page i.e. [pubsub_walkthrough](https://docs.nats.io/nats-concepts/core-nats/pubsub/pubsub_walkthrough), the first publish command has typo in `message` part of command, I think it should be "NATS MESSAGE"  instead of `nats-message-1` because below that command you have given response printed by publisher which says it as "NATS MESSAGE".

Screenshot 2024-01-31 at 12 42 46 AM

I'm first time reading about NATS and going through the docs so actually at first instance it looked like a command but after examples below that I understood that it should be message body/text.

If my understanding is wrong, then my apologies in advance but if is actually not expected, then if you can assign this issue to me and I can correct.

Thank you.

robinbowes:
Fixup a couple of doc errors:

- Add a couple of missing steps to the worked core nats pubsub example.
- Fix example message mis-match (fixes #650)

### DIFF
--- a/nats-concepts/core-nats/publish-subscribe/pubsub_walkthrough.md
+++ b/nats-concepts/core-nats/publish-subscribe/pubsub_walkthrough.md
@@ -85,9 +85,15 @@ In a new shell or command prompt session, create a new subscriber that listens o
 nats sub msg.test.new
 ```
 
+### 8. Publish another message
+
+```bash
+nats pub msg.test "NATS MESSAGE 4"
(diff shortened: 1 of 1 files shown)


gcolliso (approved):
lgtm - thanks @robinbowes

PR #674, opened 2024-05-17
Naming conventions and broken links in nats-tools

roeschter:
Triggered by a discussion on slack I made the naming restrictions and conventions more explicit. Also added that UTF-8 is deprecated.

Fixed a broken link in nats-tools (escaping _ in URLs is not required)


roeschter:
Warming up making small changes. Eventually I hope to fill some gaps and add examples in a number of places.

PR #678, opened 2024-05-30
fixed go install command for nats-top

bart1024:


gcolliso (approved):
lgtm - thanks @bart1024

PR #684, opened 2024-06-07
Update pingpong.md with stronger references to configuration terms

bannatech:
The paragraphs had loose references to the configuration terms, revised to anchor the configuration terms to the topic in each paragraph.

Reworded to capture the intention of the paragraph in the first words.

Also edited the Java example and added the C# V1 example.


gcolliso (approved):
lgtm - thanks @bannatech

PR #689, opened 2024-06-07
Wordsmithing streams.md

ScottDaniels:
Consistent use of the single quote (in two places the unicode single quote was used)

A few word changes for readability.

There is a broken link -- normally I would comment this out and tag it with FIX, but MD doesn't support comments, so I left it  (it was already marked broken, and no indication of what it was supposed to actually point to; probably should be removed).


gcolliso (approved):
lgtm - thanks @ScottDaniels

PR #690, opened 2024-06-07
Update streams.md examples

scottf:
* Edited the java example
* Added a C# V1 example
* Changed other examples to use identical stream and subject names


gcolliso (approved):
lgtm - thanks @scottf

PR #691, opened 2024-06-07
Added an intro, updated the cli screen captures with current cli output.

aricart:


gcolliso (approved):
lgtm - thanks @aricart

PR #698, opened 2024-06-12
Documented $JS.FC and $JS.ACK - Fixes sync_interval

roeschter:


gcolliso (approved):
lgtm - thanks @roeschter

PR #702, opened 2024-06-14
Minor fixes

aricart:
I believe this is a great candidate to be fully re-written. The documentation here is mostly written as a way of understanding the protocol, and while that may be valuable it is possibly more valuable to client devs than users.


gcolliso (approved):
lgtm - thanks @aricart

PR #709, opened 2024-06-24
Update anatomy.md

boblehest:
Remove duplicate sentence, which is already written 3 paragraphs up under the same heading.


gcolliso (approved):
lgtm - thanks @boblehest

PR #711, opened 2024-06-27
Fix jetstream docs on mirrors vs sources 

ioanluca:
This sentence made me think mirrors are somehow getting deprecated. Removed it to avoid confusion 

see slack conversation https://natsio.slack.com/archives/CM3T6T7JQ/p1719506018534819


gcolliso (approved):
lgtm - thanks @ioanluca

PR #715, opened 2024-07-10
Source and mirror details

roeschter:
Moved Source and Mirror to its own article. Added clarification on exepcted behavior for workqueue and interest based streams along with other not obvious behavior.


nats-concepts/jetstream/source_and_mirror.md
```diff
@@ -0,0 +1,69 @@
+# Source and Mirror
+
+When a stream is configured with a `source` or `mirror`, it will automatically and asynchronously replicate messages from the origin stream. There are several options when declaring the configuration.
+
```
jnmoyne, line 4: Though you touch on in it "Expected behavior" I would put right here at the start in this first paragraph something about the replication being "reliable store and forward" kind of replication. To make sure people get that it's reliable so all the messages are replicated (eventually) even in the face of failures (e.g. leaf nodes on partially connected devices).

nats-concepts/jetstream/source_and_mirror.md
```diff
@@ -0,0 +1,69 @@
...
+{% hint style="info" %}
+`Sources` is a generalization of the `Mirror` and allows for sourcing data from one or more streams concurrently. 
+If you require the target stream to act as a read-only replica:
+* Configure the stream without listen subjects **or**
+* Temporarily disable the listen subjects through client authorizations. 
+{% endhint %}
+
+## Source specific
+A stream defining `Sources` is a generalized replication mechanism and allows for sourcing data from **one or more streams** concurrently as well as allowing direct write/publish by clients. Essentially the source streams and client writes are aggregated into a single interleaved stream.
```
jnmoyne, line 27: You can even source the same stream more than once as long as the combination of the transform source + transform destination is unique for the stream being sourced more than once. 

(e.g. you can source foo twice, once with source "foo" and destination "bar" and once with source "foo" and destination "foobar" and indeed the message will be sourced twice and end up being 'duplicated' in the sourcing stream (though with two different subjects))

nats-concepts/jetstream/source_and_mirror.md
```diff
@@ -0,0 +1,69 @@
...
+Subject transformation and filtering allow for powerful data distribution architectures.
+
```
jnmoyne, line 29: A source can also listen to subjects at the same time (a mirror can not listen to anything, just mirror one other stream)

nats-concepts/jetstream/source_and_mirror.md
```diff
@@ -0,0 +1,69 @@
...
+{% endhint %}
+
+## Expected behavior in edge conditions
+
+* Source and mirror contracts are designed with one-way (geographic) data replication in mind. Neither configuration provides full synchronization between streams, which would include deletes or replication of other stream attributes.
+
+* The content of the stream from which a source or mirror is drawn needs to be reasonable stable. Quickly deleting messages after publishing them may result in inconsistent replication due to the asynchronous nature of the replication process.
+
+* Sources and Mirror try to be be efficient in replicating messages and are lenient towards the source/mirror origin temporarily not being reachable, e.g. when using leaf nodes, which are connected intermittently. For sake of efficiency the recovery interval in case of a disconnect is 10-20s.
```
jnmoyne, line 49: Technically it's not temporary, you could have the sourcing stream not being able to see the stream being sourced for one year and it would still recover the missed messages, the limit is more the limits of the retention of messages on the stream being sourced/mirrored (ie if the disconnect is 1 year and the stream being sourced has a max age of 1 month, after one year the most it will be able to recover is the last month).

It is also expected that when adding a source to a stream, it can take up to a few seconds for the sourcing to start (even if the stream being sourced is up and reachable).

nats-concepts/jetstream/source_and_mirror.md
```diff
@@ -0,0 +1,69 @@
...
+
+* Mirror and source agreements do not create a visible consumer in the origin stream.
+
+###  WorkQueue retention
+
+Source and mirror work with origin stream with workqueue retention. The source/mirror will act as a consumer removing messages from the origin stream. 
+
+The intended usage is for moving messages conveniently from one location to another (e.g. from a leaf node). It does not trivially implement a distributed workqueue. But a with the help of subject filtering can approximate a distributed workqueue.
+
```
jnmoyne, line 58: For your info (not sure how much we want to mention/advertise this feature very much at this point as it's working but not super-optimized in the current implementation) you can even have a working queue stream sourcing from another stream, and if that working queue has a limit _and_ a discard policy of 'new' it will work as a 'size bounded rolling buffer' type of thing - which is a pretty cool feature IMHO.

nats-concepts/jetstream/source_and_mirror.md
```diff
@@ -0,0 +1,69 @@
...
+{% hint style="warning" %}
```
jnmoyne, line 59: Would also mention that if you have a stream that sources from a WQ stream you should be careful to not have some application also try to create a consumer on that stream (if the source's consumer is active, the application will not be able to create that consumer but if the sourcing stream is down/disconnected at the time then the application would be able to successfully create that consumer, and to consume message which would then get deleted from the stream and never sourced when the sourcing stream comes back up).

jnmoyne (approved):
Besides my comments LGTM

nats-concepts/jetstream/source_and_mirror.md
roeschter: Extended description

nats-concepts/jetstream/source_and_mirror.md
roeschter: Was already mentioned, but I made it more explicit. (Redundancy is good in documentation)

nats-concepts/jetstream/source_and_mirror.md
roeschter: Explicitley descibed intermittently connected leaf nodes as a pattern

nats-concepts/jetstream/source_and_mirror.md
roeschter: We leave it out

nats-concepts/jetstream/source_and_mirror.md
roeschter: There was a warning already. I extended the description.

roeschter:
Changes complete - added example JSON as NATS CLI only allows trivial source/mirror config from command line unless a JSON input is being used. 
When I first did this 2 months back I struggled to find and example for a stream config JSON with sources

gcolliso (approved):
lgtm - thanks @roeschter & @jnmoyne

PR #716, opened 2024-07-11
Added an advantage for queue group.

BenBasilTomy:
Running consumers in multiple containers without getting duplicate messages could be achieved with queue group. Its not mentioned in the doc.


gcolliso (approved):
lgtm - thanks @BenBasilTomy

PR #717, opened 2024-07-12
Adds how to install the cli tools from the shell in Windows

jnmoyne:
Adds how to install the cli tools from the shell in Windows


gcolliso (approved):
lgtm - thanks @jnmoyne

PR #726, opened 2024-07-23
Fix example subject hierarchy

robinbowes:
Whilst doing some background reading, I noticed that the example subject hierarchy os possibly not quite right.

This PR fixes the apparent problem.


robinbowes:
Ugh - cloned from the wrong repo - closing this.

PR #731, opened 2024-07-29
Add installing using curl to natscli

jnmoyne:
Adds how to install natscli using curl from the shell to the natscli readme


gcolliso (approved):
lgtm - thanks @jnmoyne

PR #740, opened 2024-08-26
JetStream on Leaf Nodes - fixed wrong duplicated subject section in stream source example

pricelessrabbit:
In the doc section related to copy a stream from a different account and JS domain, an example is provided to build a subject with a unique pattern. 
The PR fix a typo in the example that duplicates the `importing domain` section of the subject.


gcolliso (approved):
lgtm - thanks @pricelessrabbit

PR #742, opened 2024-09-03
Fix small mistake in describing the work queue retention policy

jnmoyne:
Fix small mistake in describing the work queue retention policy


jnmoyne:
Also moved up work queue before interest as most people will use work queue over interest so more important IMHO

gcolliso (approved):
lgtm

PR #747, opened 2024-09-12
Update the nats-server version given in the example (as people will copy and paste it the first time they install nats-server to play with it) with a much better version number.

jnmoyne:


gcolliso (approved):
lgtm - thanks @jnmoyne

PR #748, opened 2024-09-13
Fix typo in server check cli command example

alin509:


gcolliso (approved):
lgtm - thanks @alin509

PR #751, opened 2024-09-23
Update kv_walkthrough.md

jk05:


alexbozhenko:
@jk05 thanks for bringing this up!

@scottf @jnmoyne @gcolliso  tagging you since you have contributed to the file:
https://github.com/nats-io/nats.docs/commits/master/using-nats/developing-with-nats/js/object.md

I feel like that documentation is doomed to be out of date, since people who change the client won't remember to update the docs here. Should we leave just links to the relevant docs, and remove the copy-paste here?
Or alternatively, if we figure out how to make github files to work here(https://www.gitbook.com/integrations/github-files) we can just point the docs to a file(or range of lines) in the corresponding client library.

PR #754, opened 2024-10-06
Fix broken link to jetstream entrypoint

yati-sagade:
By default it seems to want to go to https://github.com/nats-io/nats.docs/blob/master/nats-concepts/jetstream/README.md, which 404s.


gcolliso:
Fixed this with an internal PR for Gitbook - thank you for the notification @yati-sagade!

PR #755, opened 2024-10-09
Remove internal only subjects

johnweldon:
`$SYS.ACCOUNT.<id>.LEAFNODE.CONNECT` is marked as internal only [^1]
  `$SYS.ACCOUNT.<id>.LEAFNODE.DISCONNECT` doesn't exist

[^1]: https://github.com/nats-io/nats-server/blob/e68ac05b18eee0e2d919e47205c14b6c3072de95/server/events.go#L70


gcolliso (approved):
lgtm - thanks @johnweldon

PR #757, opened 2024-10-17
Add stream get command for workqueue

zakk616:
Workqueue streams do not support the command `nats stream view`, so added `nats stream get` command for looking into messages in the streams.


gcolliso (approved):
lgtm - thank you @zakk616

PR #769, opened 2024-12-03
Jetstream domain

roeschter:
Jetstream domain setting was missing in general. People completely miss the feature.


gcolliso (approved):
lgtm - thanks @roeschter

PR #773, opened 2024-12-04
Add a cross-link to queue permissions.

apollo13:
Couldn't find the referred server functionalities in the first place, so a cross link might make sense. Otherwise people think they miss out on important features ;)


gcolliso (approved):
lgtm - thanks @apollo13

PR #779, opened 2024-12-12
[ADDED] TLS-first handshake section

### LINKED ISSUE #778: Add documentation for `handshake_first` to `tls{}` block for client connections
The feature was added a while ago, but although this has been documented for leafnodes (https://docs.nats.io/running-a-nats-service/configuration/leafnodes/leafnode_conf#tls-block), it hasn't been added to the TLS block for client connections. Unlike for leafnodes, the parameter for client connections also support a duration, which is a fallback to revert to previous behavior of first sending the INFO in clear-text. This is to support a fleet of clients that would be at different version levels including some that do not support the "TLS handshake first" feature.

kozlovic:
This is for client connections.

Resolves #778

Signed-off-by: Ivan Kozlovic <ivan@synadia.com>

### DIFF
--- a/running-a-nats-service/configuration/securing_nats/tls.md
+++ b/running-a-nats-service/configuration/securing_nats/tls.md
@@ -54,11 +54,11 @@ When a `tls` section is specified at the root of the configuration, it also affe
 
 _As of NATS v2.10.4_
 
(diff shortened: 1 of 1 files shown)


gcolliso (approved):
thanks @kozlovic 😃

PR #780, opened 2024-12-14
Update profiling.md: add info on cli profiling

alexbozhenko:


neilalexander (commented):
Suggest we make clearer that this is subtly different to the profiling port in that `nats server request profile` by default queries *all* servers for profiles unless filtered specifically, let’s provide an example of that.

alexbozhenko:
@neilalexander good point! I added a note about `all servers`. But I do not know how to filter out specific servers.
Feel free to edit the text directly in this branch, or leave suggested text.

neilalexander:
Posted some changes, WDYT?

neilalexander (approved):
LGTM, let's ask @gcolliso to take a look.

gcolliso (approved):
lgtm - thanks @alexbozhenko @neilalexander

PR #786, opened 2024-12-28
Upd headers to include endpoint name, so it is easier to navigate

alexbozhenko:
People refer to those enpoints by their address, e.g. "Check out `/varz`", so let's include it in the headers


alexbozhenko:
https://docs.nats.io/~/revisions/dNYcwBuVZEcyD7kQitxD/running-a-nats-service/nats_admin/monitoring

PR #788, opened 2024-12-31
Update Object Store Go interfaces

HeCorr:
This PR updates the Go interfaces at [this Doc page](https://docs.nats.io/using-nats/developer/develop_jetstream/object) to their current version in [`jetstream/object.go`](https://github.com/nats-io/nats.go/blob/main/jetstream/object.go) (as of v1.38.0).

It also adds a "See more" link to the source file at the bottom, instead of bloating the page with all the other interfaces (I've only updated the existing ones).


gcolliso (approved):
lgtm - thanks @HeCorr

PR #796, opened 2025-01-10
Fix streams C example code block

evantypanski:
I unfortunately can't build the Gitbook right now (probably some `npm` weirdness with the instructions) so I can't be 100% sure this is correct, but I think it is. Right now before this change this is what the end of the code block looks like:

Screenshot 2025-01-10 at 10 52 50 AM

https://docs.nats.io/using-nats/developer/develop_jetstream/streams


gcolliso (approved):
lgtm - thanks @evantypanski

PR #801, opened 2025-01-17
Update jetstream_leafnodes.md

francoisprunier:
- Fixed the configuration for cross account stream sourcing/mirroring
- Updated the console output.

I tested the config on my machine, everything should be working as expected now.


gcolliso:
@Jarema could you review, pls?

PR #809, opened 2025-02-01
add min_version to tls.md

zarqman:
As added to nats-server in nats-io/nats-server#5904


gcolliso (approved):
lgtm - thanks @zarqman

PR #810, opened 2025-02-10
UTM Addition

maxdemaio:
This PR adds UTMs so backlinks to Synadia can be managed in a more accurate manner. This also replaces NGS with Cloud where I noted it.

cc: @connnolly


gcolliso (approved):
lgtm - thanks @maxdemaio

PR #824, opened 2025-03-27
NATS admin signals.md: document command -> signal mapping

derhuerst:
While it is documented which signals cause which server behaviour, I had to look into the source code to find out [how the commands are mapped to signals](https://github.com/nats-io/nats-server/blob/ac23d77f49a4f80f98ea6ea56f8e42d09ac8fa77/server/signal.go#L144-L162). This PR documents the mapping.


wallyqs (approved):
LGTM

gcolliso (approved):
lgtm - thanks @derhuerst

PR #828, opened 2025-04-11
Clarify that BackOff overrides AckWait in JetStream Consumer Redelivery

ndunks:
This PR updates the JetStream documentation to clarify the relationship between `AckWait` and `BackOff` in consumer configurations.

#### Background:
When configuring a JetStream consumer, the `AckWait` parameter controls the redelivery timeout **only if** `BackOff` is not set. Once `BackOff` is configured, the server ignores `AckWait` entirely and uses the provided backoff values to determine redelivery timing.

This behavior is not clearly stated in the current documentation, and can lead to confusion, especially when `AckWait` is set with the expectation that it will be respected during message redelivery.

#### Changes:
- Adds a clear note in the consumer configuration documentation that `BackOff`, when set, overrides `AckWait`.

#### Why It Matters:
Many users mistakenly expect `AckWait` to remain in effect even when using `BackOff`. Explicit documentation will help reduce confusion and prevent duplicate processing issues caused by misunderstood timing behavior.

#### Testing:
This is a documentation-only change. No code modifications or functional changes are introduced.


jnmoyne:
LGTM. Thank you very much for your contribution! It's a detail that should absolutely be documented and your changes are perfect. @gcolliso it's good to merge.

gcolliso (approved):
lgtm - thanks @ndunks

PR #833, opened 2025-05-02
Workqueue source mirror warning

roeschter:


roeschter:
We still don't think this works reliably over leaf node connections. Maurice checked the sources and confirmed that there is no resilient ack between the origin workqueue stream and the target stream

roeschter:
Yes, confirmed with JNM

PR #835, opened 2025-05-14
Update ngs-leafnodes-docker.md

elfaralfreds:
www.cloud.synadia.com does not exist, should be cloud.synadia.com


bruth (approved):
Thanks!

PR #836, opened 2025-05-17
Fix typo in `source_and_mirror.md`

escalopa:
Found 2 typos in file `nats-concepts/jetstream/source_and_mirror.md`

As Is: 

> able to consume **ad** remove messages`

To Be: 

> able to consume **and** remove messages

---

As Is:

> **will** leaf connection is down

To Be:

> **while** leaf connection is down


gcolliso (approved):
lgtm - thanks @escalopa

PR #839, opened 2025-05-22
docs: link the jetstream cross-account use case

kejne:
I had a hard time finding information about how to handle Jetstreams cross account and had not looked into mirroring more since it seemed to be for high-latency connections and similar.

Adding a simple statement to help users in the right direction.


gcolliso (approved):
lgtm - thanks @kejne

PR #840, opened 2025-05-25
fix: broken link

kejne:
Missed how the link should be created and lacked the local setup for gitbook to try it out. This should fix it, though


gcolliso:
fixed in another PR in app - thanks for the notification @kejne

PR #849, opened 2025-07-22
fix binaries.nats.dev wording

alexbozhenko:
Resolving of `@latest` was fixed in binaries.nats.dev.
Now, just like `go get`
https://go.dev/ref/mod#version-queries
`@latest` selects "the highest available release version."


gcolliso (approved):
lgtm

PR #860, opened 2025-08-21
Enrcyption at rest, TPM, cache resolver

roeschter:
Clarification on encryption at rest and copying TPM docs over. 
Cache resolver clarifications


roeschter:
Yes, its ready

PR #865, opened 2025-09-09
kv.md: add example to get an existing bucket

avivace:
It feels a bit strange that the command to retrieve an *existing* bucket is not mentioned in the Python example

Signed-off-by: Antonio Vivace <avivace4@gmail.com>


gcolliso:
@wallyqs could you review, please? TY

PR #872, opened 2025-09-25
Update example after fixing output in `stream copy`.

rafalkrupinski:
The message printed by `str cp old new` was `old was created`, while it should be `new was created`.

Reported in nats-io/natscli#1511, fixed in nats-io/natscli#1512


rafalkrupinski:
any chance for someone to have a quick look?

PR #882, opened 2025-10-24
Update authentication

roeschter:
Token authentication is not supported for leaf nodes. The examples still used it no idea why. Also added proper accounts.


gcolliso (approved):
lgtm - thanks @roeschter

PR #897, opened 2025-12-08
Update JavaScript example to use timeout instead of reconnectTimeWait

muraliprajapati:
The documentation for the connect timeout suggests using `reconnectTimeWait` option, but the correct option should be `timeout`.


gcolliso:
@aricart could you review please?

aricart (approved):
LGTM

PR #903, opened 2026-01-21
fix(subject-mapping): fix hint and code nesting

thejoeejoee:
Removed hint to fix broken nested code.

image


gcolliso (approved):
lgtm - thanks @thejoeejoee

PR #913, opened 2026-02-22
Modify user description command in basics.md

yoonghm:
Command `nsc describe user` requires an argument.


gcolliso (approved):
lgtm - thanks @yoonghm

PR #915, opened 2026-03-06
Fix invalid config for example source stream

JoSkala:
The example source stream config was invalid because of trailing commas.


gcolliso (approved):
lgtm

PR #916, opened 2026-03-09
Correct headers server versions

MauriceVanVeen:
Features were documented to be introduced in 2.11 or 2.12, but some were introduced in 2.12 (not 2.11) or are scheduled for 2.14 and are not in 2.12.


gcolliso (approved):
lgtm

PR #924, opened 2026-04-30
Add 2.14 features

MauriceVanVeen:


nats-concepts/jetstream/headers.md
sciascid: While looking at this I noticed "Monotonously",  I think the right word here is "Monotonically". Could you fix while you are at it?

MauriceVanVeen: Yep, that doesn't sound right. Also changed "increasing id" to "increasing sequence number".

sciascid (approved):
LGTM

neilalexander (approved):
LGTM

PR #929, opened 2026-05-12
UNique_tag in superclusters - reloading config for certificate changes

roeschter:
Unique_tag must be the same across all nodes in a supercluster
Making ti explicit that changing certificates will only be applied on config reload


gcolliso (approved):
lgtm - thanks @roeschter

PR #930, opened 2026-05-14
Document tls connection_rate_limit option

jarretlavallee:
Prior to this change, the parameter was not documented. This change adds some basic information about it.


gcolliso (approved):
lgtm - thanks @jarretlavallee

